PR #60, opened 2019-12-01
add support for database column types

paulstuart:
Having column type support is a handy thing for ORMs.

### DIFF
--- a/internal/protocol/message.go
+++ b/internal/protocol/message.go
@@ -503,8 +503,7 @@ type Rows struct {
 	message *Message
 }
 
-// Next returns the next row in the result set.
-func (r *Rows) Next(dest []driver.Value) error {
+func (r *Rows) columnTypes() ([]uint8, error) {
 	types := make([]uint8, len(r.Columns))
 
 	// Each column needs a 4 byte slot to store the column type. The row
@@ -522,12 +521,12 @@ func (r *Rows) Next(dest []driver.Value) error {
 
 		if slot == 0xee {
 			// More rows are available.
-			return ErrRowsPart
+			return nil, ErrRowsPart
 		}
 
 		if slot == 0xff {
 			// Rows EOF marker
-			return io.EOF
+			return nil, io.EOF
 		}
 
 		index := i * 2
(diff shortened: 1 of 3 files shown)


internal/protocol/message.go
```diff
@@ -503,8 +503,7 @@ type Rows struct {
 	message *Message
 }
 
-// Next returns the next row in the result set.
-func (r *Rows) Next(dest []driver.Value) error {
+func (r *Rows) columnTypes() ([]uint8, error) {
```
freeekanayaka: This new function reads part of the messages to decode the types of the columns of the next row. However it might be called both by ```Rows.Next()``` or by ```Rows.ColumnTypes()```: it seems that whoever call site invokes it first, will invalidate the other call site, because that part of the message will be consumed. Am I missing something?

freeekanayaka: I've added this test to ```driver/integration_test.g``` and it fails:

```
func TestIntegration_ColumnTypeName(t *testing.T) {
	db, _, cleanup := newDB(t)
	defer cleanup()

	_, err := db.Exec("CREATE TABLE test (n INT, UNIQUE (n))")
	require.NoError(t, err)

	_, err = db.Exec("INSERT INTO test (n) VALUES (1)")
	require.NoError(t, err)

	rows, err := db.Query("SELECT n FROM test")
	require.NoError(t, err)
	defer rows.Close()

	types, err := rows.ColumnTypes()
	require.NoError(t, err)

	assert.Equal(t, "INTEGER", types[0].DatabaseTypeName())

	require.True(t, rows.Next())
	var n int64
	err = rows.Scan(&n)
	require.NoError(t, err)

	assert.Equal(t, int64(1), n)
}
```

If I remove:

```
	types, err := rows.ColumnTypes()
	require.NoError(t, err)

	assert.Equal(t, "INTEGER", types[0].DatabaseTypeName())
```

then the test succeeds. I believe it's because of the issue I mentioned.

paulstuart: You are correct, that was sloppy on my part. I will resolve.

paulstuart: I've changed the columnTypes func to have a non-destructive option. All tests pass (including the one you just created).

freeekanayaka: Thanks. Please could you add the test itself to ```integration_test.go``` ? So we increase code coverage and reduce the risk of regressions.

freeekanayaka:
Thanks!

PR #63, opened 2019-12-06
cache column types to allow reporting on same at end of rows

paulstuart:
This fixes a corner case where If a client requests the column types and it is already at the last row, it will return the cached values from the previous read.

This assumes that the column types *cannot* change from row to row.

### DIFF
--- a/.dir-locals.el
+++ b/.dir-locals.el
@@ -1,7 +1,7 @@
 ;;; Directory Local Variables
 ;;; For more information see (info "(emacs) Directory Variables")
 ((go-mode
-  . ((go-test-args . "-tags libsqlite3 -timeout 10s")
+  . ((go-test-args . "-tags libsqlite3 -timeout 60s")
      (eval
       . (set
 	 (make-local-variable 'flycheck-go-build-tags)
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -18,6 +18,7 @@ import (
 	"context"
 	"database/sql/driver"
 	"io"
+	"log"
 	"net"
 	"reflect"
 	"time"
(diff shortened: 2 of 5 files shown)


freeekanayaka:
I'm confused by this change. Even after looking at it a couple of times, I don't understand what problem it solves. Please could you write a unit test that fails without this change and succeeds with it? An example is worth 1000 words :)

paulstuart:
Sure, happy to add supporting unit tests. The problem it solves is a case that shouldn't happen with better written code, but does with SQLAlchemy: it may call ColumnTypes() after the final NextI() on the rows set.

At that point it cannot get that data from the message buffer (EOF already occurred), and therefore ColumnTypes returns (nil, EOF) rather than the column types, and this bubbles up into a panic in the driver (because it can't return an error it would panic instead).

freeekanayaka:
Okay, that helps already. I'll look again at the code, still bonus points for unit tests :)

paulstuart:
Unit tests have been added. I included (but skipped) a test that checks for a column types if the query has no rows, as it fails. It appears that dqlite shortcuts the response with an EOF in that case? If this is so, would it be ok to change for "completeness"?

.dir-locals.el
```diff
@@ -1,7 +1,7 @@
 ;;; Directory Local Variables
 ;;; For more information see (info "(emacs) Directory Variables")
 ((go-mode
-  . ((go-test-args . "-tags libsqlite3 -timeout 10s")
+  . ((go-test-args . "-tags libsqlite3 -timeout 60s")
```
freeekanayaka: This change seems spurious.

freeekanayaka:
Ignore my last comment, github's diff seems busted.

PR #71, opened 2020-01-28
use go routines to have simultaneous server requests; first one wins

paulstuart:
When trying to connect to the leader of the cluster, the client needs to try every node of the cluster until it finds the leader. This patch tries all nodes simultaneously to avoid spinning on servers that don't respond.

### DIFF
--- a/client/client_test.go
+++ b/client/client_test.go
@@ -106,7 +106,7 @@ func TestClient_Transfer(t *testing.T) {
 	node1, cleanup := newNode(t)
 	defer cleanup()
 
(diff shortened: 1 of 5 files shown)


freeekanayaka:
This sounds like a reasonable idea, thanks. Unit tests are failing tho, see the Travis job log above.

paulstuart:
Dang! I'll review and remedy.

paulstuart:
There's a couple tests that still choke on this; going to close this for now and revisit later.

PR #74, opened 2020-02-03
Add driver connection options

paulstuart:
After fixing my other open PR I encountered erratic test results on my local instance, and there was a TODO on making timing configurable so I've added that functionality.

A couple of pauses were inserted in the tests because the given timeouts were not quite enough. Local testing now appears to pass consistently.

### DIFF
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -123,7 +123,7 @@ func WithAttemptTimeout(timeout time.Duration) Option {
 
 // WithRetryLimit sets the timeout for each individual connection attempt .
 //
-// If not used, the default is 5 seconds.
+// If not used, the default is 0 (unlimited retries)
 func WithRetryLimit(limit uint) Option {
 	return func(options *options) {
 		options.RetryLimit = limit
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -121,7 +121,7 @@ func WithAttemptTimeout(timeout time.Duration) Option {
 	}
 }
 
-// WithRetryLimit sets the timeout for each individual connection attempt .
+// WithRetryLimit sets the maximum number of connection retries.
 //
 // If not used, the default is 0 (unlimited retries)
 func WithRetryLimit(limit uint) Option {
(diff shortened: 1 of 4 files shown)


driver/driver.go
```diff
@@ -112,6 +112,24 @@ func WithConnectionBackoffCap(cap time.Duration) Option {
...
+// WithAttemptTimeout sets the timeout for each individual connection attempt .
+//
+// If not used, the default is 5 seconds.
+func WithAttemptTimeout(timeout time.Duration) Option {
+	return func(options *options) {
+		options.AttemptTimeout = timeout
+	}
+}
+
+// WithRetryLimit sets the timeout for each individual connection attempt .
```
freeekanayaka: This should probably be something like:

```WithRetryLimit sets the maximum number of connection retries.```

File: driver/driver.go [hunk 2]
```diff
@@ -184,20 +204,26 @@ func defaultOptions() *options {
 
 // Return a retry strategy with jittered exponential backoff, capped at the
 // given amount of time.
-func driverConnectionRetryStrategy(factor, cap time.Duration) strategy.Strategy {
+func driverConnectionRetryStrategy(factor, cap time.Duration, limit uint) []strategy.Strategy {
```
freeekanayaka: I'd suggest to rename the function to:

```driverConnectionRetryStrategies```

(plural), since it's now returning a slice.

File: driver/integration_test.go [hunk 3]
```diff
@@ -245,7 +248,20 @@ func newDB(t *testing.T, n int) (*sql.DB, []*nodeHelper, func()) {
 	require.NoError(t, store.Set(context.Background(), infos))
 
 	log := logging.Test(t)
-	driver, err := driver.New(store, driver.WithLogFunc(log))
+
+	// all driver options below (except logging) represent the default settings,
+	// and are included for code coverage
```
freeekanayaka: Please revert this to the previous version, i.e. without the various options, unless there is a value that differs from the default, so the test is closer to what users get if they don't use any tweaking.

It's okay if coverage decreases a bit.

If we want to increase coverage and exercise the new knobs, then it'd be better to add ad-hoc tests for them, showing that they actually have the desired effect (e.g. reducing the attempt timeout). However, I don't think it's necessary, so it's up to you if wish or not.

paulstuart: My commit lowered the coverage threshold so that a merge wasn't gonna happen. Could I back this out and add another "default knobs" test to make it happy?

freeekanayaka: As said, it's okay if your commit lowers the coverage. I'll merge the branch anyways, there will just a be a "meh" comment from the coverage bot, no big deal :)

paulstuart: I missed your last comment and migrated the options test to its one simple sanity check. I hope its not unreasonable. I can remove it if you wish.

PR #75, opened 2020-02-05
add logger creation function

paulstuart:
A new function to create a new customizable LogFunc, as well as an associated writer that uses the default go logger.

### DIFF
--- a/client/log.go
+++ b/client/log.go
@@ -50,7 +50,7 @@ func NewLogFunc(level LogLevel, prefix string, w io.Writer) LogFunc {
 		if l >= level {
 			// prepend the log level to the message
 			args = append([]interface{}{l.String()}, args...)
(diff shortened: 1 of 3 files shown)


freeekanayaka:
Ok, after thinking about this for a while, I would suggest to not introduce this extra functionality.

The current ```LogFunc``` type was explicitly designed to be as simple and as un-opinionated as possible, the reason being that logging is such a broad topic that no built-in solution provided by ```go-dqlite``` will ever serve all users, even the ones willing to output to stdout using Go's builtin logger might want to do it a bit differently than what's provided here.

Since writing your own ```LogFunc``` is trivial and requires very little code (as shown by this PR), I'd prefer to follow the design principle of "don't add a public API for something easy to implement" and leave ```go-dqlite``` un-opinionated about that.

PR #76, opened 2020-02-10
WIP: Add a load test

paulstuart:
This adds cmd/loadtest which will create a table and insert records into it without pause.

The table in question: `create table simple (id integer primary key, other integer, ts text)`

The canary in this coal mine is that it tracks the successful inserts and uses the returned `Result. LastInsertId` to keep track of valid the last successful insert and includes that as `other` in the record insertion.

When the number of known successful writes fails to match the last id the command will optionally stop. Examination of the table at that point will show 2 records inserted with the same value for `other`. This can be demonstrated by simultaneously running a script that transfers leadership from node to node (which I guess I should probably include?)

The errors are rare and random but do occur in time.

### DIFF
--- a/cmd/loadtest/hammer.go
+++ b/cmd/loadtest/hammer.go
@@ -44,7 +44,7 @@ func getDB(ctx context.Context, dbName string, cluster []string, logger LogFunc)
 	logOpt := driver.WithLogFunc(logger)
 	dbDriver, err := driver.New(store, logOpt)
 	if err != nil {
-		return nil, fmt.Errorf("failed to create dqlite driver: %w", err)
+		return nil, err
 	}
 	sql.Register("dqlite", dbDriver)
 
--- a/cmd/loadtest/hammer.go
+++ b/cmd/loadtest/hammer.go
@@ -1,0 +1,141 @@
+package main
+
+import (
+	"context"
+	"database/sql"
+	"fmt"
+	"log"
+	"time"
+
+	"github.com/canonical/go-dqlite/client"
+	"github.com/canonical/go-dqlite/driver"
+)
+
+// match log formatting with Lmicroseconds set
+const timeLayout = "2006/01/02 03:04:05.000000PM -0700"
+
+func ts() string {
+	return time.Now().Local().Format(timeLayout)
+}
+
+type tsWriter struct{}
+
+func (t tsWriter) Write(in []byte) (int, error) {
+	fmt.Println(ts(), string(in))
+	return len(in), nil
+}
+
+func getStore(ctx context.Context, cluster []string) client.NodeStore {
+	store := client.NewInmemNodeStore()
+	infos := make([]client.NodeInfo, len(cluster))
+	for i, address := range cluster {
+		infos[i].ID = uint64(i + 1)
+		infos[i].Address = address
+	}
+	store.Set(ctx, infos)
+	return store
+}
+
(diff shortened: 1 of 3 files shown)


freeekanayaka:
Thanks, this seems like a good first step.

I suppose that the program is intended to be run against the cluster created with the demo program?

How did you trigger the leadership transfers during your test runs? By hand or do you have something automated (like the script you mention)?

The more automated and repeatable we can make this, the easier will be to debug.

As I mentioned before, I'm well aware that there some required behavior (i.e. the client retrying failed transactions, and server handling duplicate retries) is not implemented yet, and most probably that is the cause of this issue.

paulstuart:
First and foremost, I do recognize that you’re aware of the issue and that your plate is full, and  that you are doing what you can. I’m highly motivated to see this addressed and for the project as a whole to thrive. I want to help to the best of my abilities and minimize my “bothering you”. I could start by being better about polishing my PRs ;-).

The automation part is understood as well, but the only way that has come to mind to possibly recreate the issue is to add a sleep function to sqlite to create a manageable extended transaction to create the opportunity to munge the transaction. Even then I’m unsure if that will expose the edge case in question.

I have a mutated version of dqlite-demo that adds functionality that I need for my own project, but I believe it would be better to have it go back into this project (yay open source!).

I have the following additional commands I would love to submit back:

  adhoc       execute an arbitrary SQL statement against the database.
  assign      assign a role to a node (voter, spare, or standby).
  dump       dump the database (and its associated WAL file).
  leader      print the id of the current leader node.
  load         execute the statements in the given file.
  remove    remove a node.
  transfer    transfer leadership to a new node.

I’ll add these back individually (starting with “transfer” for this issue) and see if I can hew to my goal of sterling PRs.

freeekanayaka:
> The automation part is understood as well, but the only way that has come to mind to possibly recreate the issue is to add a sleep function to sqlite to create a manageable extended transaction to create the opportunity to munge the transaction. Even then I’m unsure if that will expose the edge case in question.

I'm not entirely sure a sleep would help, since the dqlite engine is single threaded, and a sleep would just block the thread. But maybe I'm misunderstanding.

> 
> I have a mutated version of dqlite-demo that adds functionality that I need for my own project, but I believe it would be better to have it go back into this project (yay open source!).
> 
> I have the following additional commands I would love to submit back:
> 
> adhoc execute an arbitrary SQL statement against the database.
> assign assign a role to a node (voter, spare, or standby).
> dump dump the database (and its associated WAL file).
> leader print the id of the current leader node.
> load execute the statements in the given file.
> remove remove a node.
> transfer transfer leadership to a new node.
> 
> I’ll add these back individually (starting with “transfer” for this issue) and see if I can hew to my goal of sterling PRs.

The purpose of the demo program is mainly for people to look at the code and understand how the various pieces fit together, so they can do that in their own program. I'm slightly concerned that the demo program grows beyond that scope.

Why do you need these features in the demo program? The ```client.Client``` API already offers most of that, and they are ready to be consumed in your own program. The demo program is not intended to be used in any real-world scenario.

paulstuart:
The sleep idea was a stab in the dark and can be discarded.

It wasn't my intention to bloat the demo, I just saw it as extending it to being immediately usable by others in its own right. I needed to expose the Transfer function for scripting the loadtest and I got excited about "helping more".

Please forgive my enthusiasm. I'll update the `loadtest` command to do node transfers as well to minimize changes to the codebase.

PR #77, opened 2020-02-11
add context to getLeader function

paulstuart:
Exposing context to allow timeouts to be configurable from command line functions.

### DIFF
--- a/cmd/dqlite-demo/add.go
+++ b/cmd/dqlite-demo/add.go
@@ -36,14 +36,14 @@ func newAdd() *cobra.Command {
 				Address: address,
 			}
 
-			client, err := getLeader(*cluster)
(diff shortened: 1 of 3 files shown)


freeekanayaka:
Looks reasonable to me. Please can you also extend this PR with the follow-up commits you have or intend to write in order to actually  expose these timeouts from the command line? So the branch will have the full story, I can see the whole picture and we commit it in one go. Thanks.

paulstuart:
Happy to do so, I was trying to make my offerings more chunk-sized for ease of reasoning/merging/etc.

The timeouts were to be as part of the additional commands I was hoping to add to dqlite-demo in PR #76 but that may be now moot. I'm not sure how to progress further here and if you just want to take it as is or otherwise. I defer to your decision.

PR #84, opened 2020-03-24
Use a single buffer for sending and receiving messages

freeekanayaka:
The previous implementation was using 2 buffers for sending and receiving messages: a one-off allocated buffer of fixed size reused across messages (to avoid re-allocation) and a dynamically sized one allocated for sending or receiving specific messages exceeding the size of the one-off buffer.

That was an optimization that probably didn't buy much performance, and apparently was the source of some bugs surfaced in k8s.

This branch moves to a single buffer, which is grown as needed and not re-allocated. This should maintain the performance benefits of the previous approach (however small, this would need to be measured), but simplifies the underlying logic making it more robust. The only downside is that this buffer is never shrunk once grown, if that becomes a problem in practice we can put in place some mechanism to shrink it.

### DIFF
--- a/internal/protocol/request.go
+++ b/internal/protocol/request.go
@@ -7,7 +7,7 @@ package protocol
 
 // EncodeLeader encodes a Leader request.
 func EncodeLeader(request *Message) {
-    request.Reset()
+	request.reset()
 	request.putUint64(0)
 
 	request.putHeader(RequestLeader)
@@ -15,7 +15,7 @@ func EncodeLeader(request *Message) {
 
 // EncodeClient encodes a Client request.
 func EncodeClient(request *Message, id uint64) {
-    request.Reset()
+	request.reset()
 	request.putUint64(id)
 
 	request.putHeader(RequestClient)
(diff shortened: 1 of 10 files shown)


internal/protocol/request.go
```diff
@@ -7,27 +7,31 @@ package protocol
 
 // EncodeLeader encodes a Leader request.
 func EncodeLeader(request *Message) {
+    request.Reset()
```
stgraber: Is it just github looking confused or is there really something weird going on here with tab alignment?

internal/protocol/request.go
```diff
@@ -7,27 +7,31 @@ package protocol
 
 // EncodeLeader encodes a Leader request.
 func EncodeLeader(request *Message) {
+    request.reset()
```
stgraber: Yeah, it's the only file that shows this, suggests mixed tabs/spaces.

internal/protocol/request.go
freeekanayaka: Ah it's because this code is automatically generated with ```go generate``` (see ```schema.sh```), and I used spaces instead of tab for indentation in the generation script.

Fixed now.

PR #91, opened 2020-04-30
app: choose snap suitable socket name to satisfy AppArmor confinement

morphis:
Within snaps the name of the choosen Unix domain socket has to be
special crafted. See https://github.com/snapcore/snapd/blob/master/interfaces/apparmor/template.go#L357
for more details


freeekanayaka:
I'm going to merge this as it is for now. If in the future we discover more use cases where customizing the socket name is needed, we might generalize this a bit and actually expose an option (e.g. ```app.WithUnixSocket``` or whatever sensible name) that lets you explicitly specify the socket name.

PR #94, opened 2020-05-11
App & client: set tls.Config.ServerName for every connection

aguadoenzo:
Because of https://github.com/FiloSottile/go/commit/268080 ServerName
needs to be set in the TLS config in order to Write() directly to a TLS
connection. This change tasks dialers to set the proper ServerName when
making a connection rather than setting it once for every connection

This lets each dqlite node use its own certificate when running in a cluster
rather than forcing the same one to be used.

The certificates are changed to have their SAN IP set to 127.0.0.1 as this is
the address used in unit tests


client/dial.go
```diff
@@ -24,10 +24,16 @@ func DefaultDialFunc(ctx context.Context, address string) (net.Conn, error) {
 // and the given TLS config will be used for encryption.
 func DialFuncWithTLS(dial DialFunc, config *tls.Config) DialFunc {
 	return func(ctx context.Context, addr string) (net.Conn, error) {
+		clonedConfig := config.Clone()
+		remoteIP, _, err := net.SplitHostPort(addr)
+		if err != nil {
+			return nil, err
+		}
+		clonedConfig.ServerName = remoteIP
```
freeekanayaka, line 32: This makes it possible for each node to have its own certificate, however it breaks the current "simple" setup where a single certificate gets generated on a node and then shared with all the others.

What do you think of changing this to:

```golang
if clonedConfig.ServerName == "" {
		remoteIP, _, err := net.SplitHostPort(addr)
		if err != nil {
			return nil, err
		}
		clonedConfig.ServerName = remoteIP
}
```

?

app/tls.go
```diff
@@ -108,7 +108,6 @@ func SimpleDialTLSConfig(cert tls.Certificate, pool *x509.CertPool) *tls.Config
 	if len(x509cert.DNSNames) == 0 {
 		panic("certificate has no DNS extension")
 	}
-	config.ServerName = x509cert.DNSNames[0]
```
freeekanayaka, line 111: I think this one should be kept, since the goal of the "simple" TLS configuration is to support sharing the same certificate across nodes. If a user wants per-node certificates it should not use these helpers and build his own `*tls.Config` object from scratch.

app/app_test.go
```diff
@@ -38,7 +38,7 @@ func TestNew_PristineJoiner(t *testing.T) {
 
 // Open a database on a fresh one-node cluster.
 func TestOpen(t *testing.T) {
-	app, cleanup := newApp(t)
+	app, cleanup := newApp(t, app.WithAddress("127.0.0.1:9001"))
```
freeekanayaka, line 41: If the other comments get addressed, this change could be reverted as well, so we test with the default which is to pick the first non-loopback address on the host.

PR #108, opened 2020-06-08
Don't join as Voter if there are less than 3 nodes

freeekanayaka:
This is an initial stab at including in the `app` framework the same goodness we have in LXD and microk8s. Joining nodes will be assigned the `Voter` role only once the cluster reaches 3 nodes. More refinements will follow, as well as handling of edge cases.


freeekanayaka:
Travis failure seems due to an unrelated flaky test. I will investigate that separately.

PR #110, opened 2020-06-19
Dynamic roles management

freeekanayaka:
This branch adds to the `app` package the same roles management goodness that LXD and microk8s currently enjoy.


freeekanayaka:
This will probably need a bit more polish, but it's a reasonable start.

PR #125, opened 2020-12-04
proxy: make socketpair fds close-on-exec

brauner:
Signed-off-by: Christian Brauner <christian.brauner@ubuntu.com>

PR #126, opened 2020-12-07
proxy: close remote

brauner:
In all other paths we simply call remote.Close and local.Close but in
this specific part we call CloseRead():
- I would just like to know why we use CloseRead() and not Close()
- The proxy() function is called from

```go
func (a *App) proxy() {
	wg := sync.WaitGroup{}
	ctx, cancel := context.WithCancel(context.Background())
	for {
		client, err := a.listener.Accept()
		if err != nil {
			cancel()
			wg.Wait()
			close(a.proxyCh)
			return
		}
		address := client.RemoteAddr()
		a.debug("new connection from %s", address)
		server, err := net.Dial("unix", a.nodeBindAddress)
		if err != nil {
			a.error("dial local node: %v", err)
			client.Close()
			continue
		}
		wg.Add(1)
		go func() {
			defer wg.Done()
			if err := proxy(ctx, client, server, a.tls.Listen); err != nil {
				a.error("proxy: %v", err)
			}
		}()
	}
}
```

and it seems to me that by not calling close on client (named "remote"
in the proxy function) we might be leaking a connection? Or am I
misunderstandig this. Because it doesn't look like client is closed
properly in this scenario. But maybe I have this all backwards!

Signed-off-by: Christian Brauner <christian.brauner@ubuntu.com>

### DIFF
--- a/app/proxy.go
+++ b/app/proxy.go
@@ -79,6 +79,7 @@ func proxy(ctx context.Context, remote net.Conn, local net.Conn, config *tls.Con
 		if err := <-remoteToLocal; err != nil {
 			errs[1] = fmt.Errorf("remote -> local: %v", err)
 		}
+		remote.Close()
 		local.Close()
 
 	}


freeekanayaka:
> In all other paths we simply call remote.Close and local.Close but in
> this specific part we call CloseRead():

I assume that by "all other code paths" you mean the other branches of the select statement. Those capture different error modes that require different handling. Namely:

- The `ctx.Done()` case means that higher-level code is brute-forcing the proxy to finish, so we brute-force close everything and ignore errors.

- The `err := <-remoteToLocal` case is somehow specular to the `err := <-localToRemote` case modified by this PR, which I describe in detail just below.

>     * I would just like to know why we use CloseRead() and not Close()
>

The "remote" connection object here is the server-side end of a TLS connection established by the dqlite client when it connects to the server. The "local" connection object is the Go side of a Unix socket connection established by the `App.proxy()` method when it connects to the Unix socket that the dqlite C thread is listening to.

We stop proxying as soon as either `io.Copy(local, remote)` or `io.Copy(remote, local)` returns an error. In particular, the code modified by this PR is triggered when `io.Copy(remote, local)` returns an error **before** `io.Copy(local, remote)`  does. That case should happen only when the C dqlite loop unilaterally closes the client connection, typically because the dqlite engine is being shut down (in all other cases it should be the client that closes the connection, triggering an `io.Copy(local, remote)` error).

So when `io.Copy(remote, local)` returns an error we end up here:

```Go
	case err := <-localToRemote:
		if err != nil {
			errs[0] = fmt.Errorf("local -> remote: %v", err)
		}
		tcp.CloseRead()
		if err := <-remoteToLocal; err != nil {
			errs[1] = fmt.Errorf("remote -> local: %v", err)
		}
		local.Close()
```

Calling `tcp.CloseRead()` (where `tcp` is the underlying TCP connection object wrapped by the `remote` TLS connection object) will make the `io.Copy(local, remote)` call in the other proxy goroutine return gracefully (without errors) so the call to `err :=  <-remoteToLocal` just below will also return no error and the `proxy` function will return no error as well (assuming that the `io.Copy(remote, local)` call didn't return any error, as it should since the C dqlite thread should close that connection gracefully).

If we call `remote.Close()` instead of  `tcp.CloseRead()`, then the call to `io.Copy(local, remote)` will not terminate gracefully and return an error instead, which will be captured by `err :=  <-remoteToLocal`, returned by the `proxy()` function and logged by the `App.proxy()` method which calls `proxy()`. Since that error is a normal "error" due to the node being shut down, we should not log it, as it'd be just noise.

The `remote` connection object should be eventually garbage collected and the underlying TCP connection closed by the finalizer, resulting in no leak.

This is what should happen in theory, if in practice we notice there's actually a leak, we can certainly merge this PR, and the only side effect that it should have is to include spurious errors in the log, which would be a bit of a noise (perhaps leading users to think there's something wrong going on), but not harmful.

Alternatively I believe we could mix the two approaches and **also** call `remote.Close()`, right after the call to `local.Close()` in the `err := <-localToRemote` select branch.

app/proxy.go
```diff
@@ -75,10 +75,10 @@ func proxy(ctx context.Context, remote net.Conn, local net.Conn, config *tls.Con
 		if err != nil {
 			errs[0] = fmt.Errorf("local -> remote: %v", err)
 		}
-		tcp.CloseRead()
```
freeekanayaka, line 78: I'd suggest to change this branch and KEEP this line that is being deleted, as explained it should avoid logging spurious errors when the dqlite engine is being shutdown and client connections get legitimately closed.

The `remote.Close()` call that this branch adds just below should be fine and will serve as extra safety net.

brauner: Done!

freeekanayaka (approved):

freeekanayaka:
Looks good to me, we can merge it as soon as the Travis integration is sorted.

brauner:
> Looks good to me, we can merge it as soon as the Travis integration is sorted.

As things look right now that's going to be never.

brauner:
> > Looks good to me, we can merge it as soon as the Travis integration is sorted.
> 
> As things look right now that's going to be never.

See @stgraber's comment here:
https://github.com/canonical/go-dqlite/pull/125#issuecomment-740254416

brauner:
> Okay, merging this particular one anyways since it seems safe.

It's a bit annoying and if this doesn't get better fast we need to migrate to something else...

freeekanayaka:
FWIW I've ran the tests locally against master with this branch merged and they were green.

PR #132, opened 2021-02-20
Build on MacOS with TLS support

rabits:
Found a couple of ways to make the logic cross-platform in golang, so seems the patch is actually ready to be a part of the master branch.

Overall the change allows to build demo & deps (with patches to the latest releases) on MacOS and connect to the other nodes (Lin/Mac) using TLS.

To make the TLS work properly on Mac - it uses the regular socket file (abstract one are supported on Linux only) and it's created in the CWD of the process. I think it's the best way to implement, since we can't use `/tmp` or `/var/run` on Windows later.

* MacOS -> MacOS - works well
* Linux -> MacOS - works well
* MacOS -> Linux - joined and looks like the demo is working well

The other patches in series:
* canonical/dqlite#285
* https://github.com/canonical/dqlite/pull/288
* https://github.com/canonical/dqlite/pull/290
* canonical/raft#173

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -7,6 +7,7 @@ import (
 	"net"
 	"os"
 	"path/filepath"
+	"runtime"
 	"sync"
 	"time"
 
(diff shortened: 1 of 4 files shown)


stgraber:
@rabits app/proxy_linux.go:3:1: +build comment must appear before package clause and be followed by a blank line

rabits:
@stgraber yep, fixed)

PR #136, opened 2021-04-30
cli: Add .dump command

MathieuBordere:

### DIFF
--- a/internal/shell/shell.go
+++ b/internal/shell/shell.go
@@ -6,6 +6,9 @@ import (
 	"database/sql"
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
+	"os"
+	"path/filepath"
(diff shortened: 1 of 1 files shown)


MathieuBordere:
@freeekanayaka Were the CI tests flaky in the past? I don't really see how I could have broken them.

freeekanayaka:
> @freeekanayaka Were the CI tests flaky in the past? I don't really see how I could have broken them, or dqlite itself might be broken.

I think I saw some flakiness in the past, and fixed it, not sure either if this is something new or not. It's totally possible that the failure/flakiness is rare and you hit it by chance. Anyway, I agree that it's most probably unrelated to you changes.

freeekanayaka (approved):
LGTM

PR #138, opened 2021-05-28
github_actions: Translate .travis.yml to actions

MathieuBordere:
Succeeded here on my repo: https://github.com/MathieuBordere/go-dqlite/pull/2


.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,44 @@
...
+jobs:
+  build-and-test:
+    runs-on: ubuntu-18.04
+    steps:
+    - uses: actions/checkout@v2
+
+    - name: Set up Go
+      uses: actions/setup-go@v2
+      with:
+        go-version: 1.14
```
stgraber, line 20: Should test similar to go-lxc (1.13.x through 1.16.x)

.github/workflows/build-and-test.yml
```diff
@@ -0,0 +1,44 @@
+name: go-dqlite-build-and-test
```
stgraber, line 1: A more friendly name would be nice

.github/workflows/cla-check.yml
```diff
@@ -0,0 +1,10 @@
+name: go-dqlite-cla-check
+
+on: [pull_request]
```
stgraber, line 3: I'd still write it as:
```
on:
 - pull_request
```

Just for consistency

.github/workflows/cla-check.yml
```diff
@@ -0,0 +1,10 @@
+name: go-dqlite-cla-check
```
stgraber, line 1: A better name would help

stgraber: Maybe something like `Canonical CLA`?

stgraber:
Left a few comments. This should be pretty close to what we do for `go-lxc` so maybe take a look at those and try to line things up?

MathieuBordere:
> Left a few comments. This should be pretty close to what we do for `go-lxc` so maybe take a look at those and try to line things up?

Thanks, will do.

PR #142, opened 2021-06-02
LookupHostb4SetBindAddress

dacleyra:
I have a use case where a go-dqlite cluster is run as a StatefulSet on Kubernetes.

A StatefulSet has a stable FQDN, stable storage, but not necessarily a stable IP address.

Therefore it is desirable to use the FQDN in the cluster configuration, rather than the IP Address.

As is, 
app will accept a FQDN for configuration
node SetBindAddress only accepts an IP address

As such, attempt a net.LookupHost before passing to SetBindAddress

### DIFF
--- a/node.go
+++ b/node.go
@@ -1,6 +1,7 @@
 package dqlite
 
 import (
+	"net"
 	"time"
 
 	"github.com/canonical/go-dqlite/client"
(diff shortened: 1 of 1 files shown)


stgraber:
Hi Daniel,

Looks like this change is causing a few regressions in our tests (see CI test results).

As dqlite is a project owned by Canonical, we're required to have contributors sign Canonical's contributor licence agreement prior to contributing to one of the company owned project. You can sign it using your Github account at https://ubuntu.com/legal/contributors. Worth noting that this is a one time thing to contribute to any of Canonical's projects, not a per-project or per-contribution thing.

If you don't want to go through that, we totally understand. In which case the best approach would be to file an issue instead so we can get it resolved ourselves instead.

Thanks!

Stéphane

dacleyra:
it appears fqdn is used for cluster configuration once TLS is enabled, so this solution was ultimately unnecessary, and otherwise problematic

PR #143, opened 2021-06-11
app: node: Add raft snapshot parameter options

MathieuBordere:
This PR uses the new dqlite API `dqlite_node_set_snapshot_params` available from dqlite `1.8.0`.

I want to update the SONAME of dqlite from `libdqlite.so.0` to `libdqlite.so.1` so that go-dqlite binaries built against a version of dqlite that contain this API fail when ran against an older version of the dqlite library. The problem is, I already released a version of libdqlite with SONAME `libdqlite.so.0` that contains the `dqlite_node_set_snapshot_params` functionality.

I'm not entirely sure if I should:
a) make a new dqlite release `1.9.0` and update the SONAME ? <-- This probably seems most reasonable
b) just update the SONAME on master and not make a release ?
c) ... ?

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -35,6 +35,8 @@ jobs:
         go get github.com/go-playground/overalls
 
     - name: Build & Test
+      env:
+        CGO_LDFLAGS_ALLOW: "-Wl,-z,now"
       run: |
         go get -t -tags libsqlite3 ./...
         go vet -tags libsqlite3 ./...
--- a/README.md
+++ b/README.md
@@ -45,6 +45,7 @@ system, along with its dependencies. You then need to pass the ```-tags```
 argument to the Go tools when building or testing your packages, for example:
 
 ```bash
+export CGO_LDFLAGS_ALLOW="-Wl,-z,now"
 go build -tags libsqlite3
 go test -tags libsqlite3
 ```
--- a/app/app.go
+++ b/app/app.go
@@ -167,6 +167,7 @@ func New(dir string, options ...Option) (app *App, err error) {
 		dqlite.WithDialFunc(nodeDial),
 		dqlite.WithFailureDomain(o.FailureDomain),
 		dqlite.WithNetworkLatency(o.NetworkLatency),
+		dqlite.WithSnapshotParams(o.SnapshotParams),
 	)
 	if err != nil {
 		return nil, fmt.Errorf("create node: %w", err)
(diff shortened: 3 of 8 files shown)


stgraber:
@MathieuBordere hmm, to me moving from `libdqlite.so.0` to `libdqlite.so.1` would normally be the sign of an API break, not simple symbol addition.

Unless we need to move to `libdqlite.so.1` so that SEMVER works properly and then lets us do `libdqlite.so.1.1.0`, `libdqlite.so.1.2.0`, ... in the future and the same isn't currently true with `libdqlite.so.0`.

MathieuBordere:
> @MathieuBordere hmm, to me moving from `libdqlite.so.0` to `libdqlite.so.1` would normally be the sign of an API break, not simple symbol addition.
> 
> Unless we need to move to `libdqlite.so.1` so that SEMVER works properly and then lets us do `libdqlite.so.1.1.0`, `libdqlite.so.1.2.0`, ... in the future and the same isn't currently true with `libdqlite.so.0`.

I'll experiment some more with it, I need `libtool` to somehow spit out a `libdqlite.so.0.1` SONAME for it to work properly I think.

MathieuBordere:
@stgraber So there's the `-Wl,-z,relro,-z,now` linker option to fail the loading of a binary containing unknown symbols, but go only allows setting them when explicitly allowed through the environment variable `CGO_LDFLAGS_ALLOW`, so `go get` will fail when users haven't set this environment variable.

There's another alternative, [link](https://stackoverflow.com/a/67973371/3613657) but that doesn't seem to be available on all platforms, so some more configuration magic would be needed in libraries that want to use this and it won't work immediately, it will work once 2 versions of the library have applied this method.

There's no clear winning alternative for me, what would you prefer?

stgraber:
@MathieuBordere I think if we can tweak the Makefile in lxd that may be good enough.

MathieuBordere:
Let me have a look at those `TestIntegration_ExecBindError` failures, they only started coming up recently.

MathieuBordere:
> @MathieuBordere I think if we can tweak the Makefile in lxd that may be good enough.

Adapting `export CGO_LDFLAGS_ALLOW="-Wl,-wrap,pthread_create"` to something like export `CGO_LDFLAGS_ALLOW="(-Wl,-wrap,pthread_create)|(-Wl,-z,now)"` should do the trick normally.

MathieuBordere:
`test/roles.sh` is hanging a bit more than expected, want to make sure it's not related to this change in any way.

PR #146, opened 2021-06-21
CI test fixes

MathieuBordere:
Fixes a couple of "infinite" hangs and flaky unit tests.


MathieuBordere:
There's still another more rare hang apparently ... will hunt it down.

Issue #145, opened 2021-06-16
cluster.yaml becomes empty / corrupted


MathieuBordere:
see https://github.com/ubuntu/microk8s/issues/2324 & https://github.com/ubuntu/microk8s/issues/2329

Possibly related to this return value not getting checked https://github.com/canonical/go-dqlite/blob/5cc7c73154d37ae8d01f5dab8cf79608edea0e15/app/app.go#L491

PR #147, opened 2021-06-24
More robust filewriting

MathieuBordere:
There are some mentions of corrupted / empty `cluster.yaml` files. This PR should help prevent these cases.

- Atomically replace files when writing by using the [renameio](https://github.com/google/renameio) package.
- Protect against empty server list (if an empty list is ever even possible).

Fixes https://github.com/canonical/go-dqlite/issues/145

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -489,6 +489,11 @@ func (a *App) run(ctx context.Context, frequency time.Duration, join bool) {
 				cli.Close()
 				continue
 			}
+			if len(servers) == 0 {
+				a.warn("server list empty")
+				cli.Close()
+				continue
+			}
 			a.store.Set(ctx, servers)
 
 			// If we are starting up, let's see if we should
--- a/app/files.go
+++ b/app/files.go
@@ -7,6 +7,7 @@ import (
 	"path/filepath"
 
 	"github.com/ghodss/yaml"
+	"github.com/google/renameio"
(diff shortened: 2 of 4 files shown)


freeekanayaka (approved):
Looks good to me. Beware that adding a new dependency means having to update the Launchpad Debian package receipt (and import that new dependency in the repository).

MathieuBordere:
> Looks good to me. Beware that adding a new dependency means having to update the Launchpad Debian package recipe (and import that new dependency in the repository).

Thanks for the heads up, I'll have a look.

MathieuBordere:
@freeekanayaka So I think I have to update the recipe here https://code.launchpad.net/~dqlite/+recipe/dqlite, but I don't really find where I can add the repository to the list of `other repositories` here https://code.launchpad.net/dqlite.

Thanks a lot if you could give me a pointer in the right direction.

freeekanayaka:
> @freeekanayaka So I think I have to update the recipe here https://code.launchpad.net/~dqlite/+recipe/dqlite, but I don't really find where I can add the repository to the list of `other repositories` here https://code.launchpad.net/dqlite.
> 
> Thanks a lot if you could give me a pointer in the right direction.

You should be able to create a new import from GitHub using this link https://code.launchpad.net/+code-imports/+new

PR #150, opened 2021-07-20
shell: Speedup row printing

MathieuBordere:
Looks like the converting the row to a string takes a large amount of time, these changes amount to approx. a 4x speed increase when performing `select * from kine LIMIT 10000;` on a large kine database. Expect the speedup to be similar for other use cases.

### DIFF
--- a/internal/shell/shell.go
+++ b/internal/shell/shell.go
@@ -269,7 +269,7 @@ func (s *Shell) processSelect(ctx context.Context, line string) (string, error) 
 	}
 	n := len(columns)
 
-	result := ""
+	var sb strings.Builder
 	for rows.Next() {
 		row := make([]interface{}, n)
 		rowPointers := make([]interface{}, n)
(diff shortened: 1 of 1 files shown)


tomponline (approved):
Out of interest, did you also try using StringBuilder?

https://www.calhoun.io/concatenating-and-building-strings-in-go/

MathieuBordere:
> Out of interest, did you also try using StringBuilder, which would avoid the intermediate slice?
> 
> https://www.calhoun.io/concatenating-and-building-strings-in-go/

No, will give it a spin, thanks

MathieuBordere:
> Out of interest, did you also try using StringBuilder, which would avoid the intermediate slice?
> 
> https://www.calhoun.io/concatenating-and-building-strings-in-go/

Thanks, turns out using strings.Builders amounts to another 30x speedup on top of the 4x :-)

tomponline:
> > Out of interest, did you also try using StringBuilder, which would avoid the intermediate slice?
> > https://www.calhoun.io/concatenating-and-building-strings-in-go/
> 
> Thanks, turns out using strings.Builders amounts to another 30x speedup on top of the 4x :-)

Awesome! :)

tomponline:
@MathieuBordere does test coverage check failure block merges in this project?

MathieuBordere:
> @MathieuBordere does test coverage check failure block merges in this project?

It doesn't, no.

PR #151, opened 2021-07-22
Recover

MathieuBordere:
introduces a `.reconfigure` command to force a new config into the raft cluster.

@freeekanayaka Does the help section of the command look sensible/complete to you?

edit: Just found out the node roles aren't actually taken into account in the `dqlite_node_recover` command, will put on draft for now.


freeekanayaka:
The help description looks good to me. Perhaps the only thing that is not very clear is that a single node should be picked as "most up to date", then reconfigured and then its data copied over to all other nodes. Probably the whole process should be described a bit more in detail.

stgraber:
@masnax that RecoverExt function is what you'll need to use to reconfigure dqlite in `lxd cluster edit`.

PR #170, opened 2021-11-17
Tests: Replace RSA cert by ECDSA cert

MathieuBordere:
Because of https://github.com/golang/go/issues/20058

Speeds up the tests on armhf machines by quite a lot and lowers the chance of them timing out.

key & cert generation:
```bash
$ openssl version
OpenSSL 1.1.1f  31 Mar 2020
$ openssl ecparam -out cluster.key -name secp256r1 -genkey
$ openssl req -new -key cluster.key -x509 -nodes -days 10000 -out \
cluster.crt -addext "subjectAltName = IP.1:127.0.0.1, DNS:local.test"
You are about to be asked to enter information that will be incorporated
into your certificate request.
What you are about to enter is what is called a Distinguished Name or a
DN.
There are quite a few fields but you can leave some blank
For some fields there will be a default value,
If you enter '.', the field will be left blank.
-----
Country Name (2 letter code) [AU]:.
State or Province Name (full name) [Some-State]:.
Locality Name (eg, city) []:.
Organization Name (eg, company) [Internet Widgits Pty Ltd]:.
Organizational Unit Name (eg, section) []:.
Common Name (e.g. server FQDN or YOUR name) []:dqlite-test
Email Address []:.
```

Also replaced the ppa in the actions.


freeekanayaka (approved):

stgraber:
Ah yeah, that makes a lot of sense, extremely poor performance of RSA in Go on ARM is why we switched to ECDSA in LXD in the first place (at Google's request due to Chromebooks being extremely slow).

PR #179, opened 2022-02-01
Add WithExternalConn to app.App

masnax:
This takes a bunch of the logic from the dqlite proxies in lxd's `Gateway` and applies them to `app.App` via a `WithExternalConn` option.

`WithExternalConn` takes a `client.DialFunc` and an `acceptCh` which will be used for back and forth handling of the external connection. 

I modified the TLS keep alive options a little to look more like how we handle them in LXD [here.](https://github.com/lxc/lxd/blob/ddc19d45afd6344fa21d2db0a6e0fbf31c50a36c/lxd/util/net.go#L333-L376)

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -32,6 +32,7 @@ type App struct {
 	nodeBindAddress string
 	listener        net.Listener
 	tls             *tlsSetup
+	dialFunc        client.DialFunc
 	store           client.NodeStore
 	driver          *driver.Driver
 	driverName      string
@@ -50,6 +51,17 @@ func New(dir string, options ...Option) (app *App, err error) {
 	o := defaultOptions()
 	for _, option := range options {
 		option(o)
+	}
+
+	var nodeBindAddress string
+	if o.Conn != nil {
+		listener, err := net.Listen("unix", "")
+		if err != nil {
+			return nil, fmt.Errorf("Failed to autobind unix socket: %w", err)
+		}
+
+		nodeBindAddress = listener.Addr().String()
+		listener.Close()
 	}
 
 	// List of cleanup functions to run in case of errors.
@@ -147,9 +159,10 @@ func New(dir string, options ...Option) (app *App, err error) {
 	}
 
 	// Start the local dqlite engine.
-	var nodeBindAddress string
 	var nodeDial client.DialFunc
-	if o.TLS != nil {
+	if o.Conn != nil {
+		nodeDial = extDialFuncWithProxy(o.Conn.dialFunc)
+	} else if o.TLS != nil {
 		nodeBindAddress = fmt.Sprintf("@dqlite-%d", info.ID)
 
 		// Within a snap we need to choose a different name for the abstract unix domain
@@ -185,6 +198,8 @@ func New(dir string, options ...Option) (app *App, err error) {
 	driverDial := client.DefaultDialFunc
 	if o.TLS != nil {
 		driverDial = client.DialFuncWithTLS(driverDial, o.TLS.Dial)
+	} else if o.Conn != nil {
+		driverDial = o.Conn.dialFunc
 	}
 
 	driver, err := driver.New(store, driver.WithDialFunc(driverDial), driver.WithLogFunc(o.Log))
@@ -217,6 +232,7 @@ func New(dir string, options ...Option) (app *App, err error) {
 		node:            node,
 		nodeBindAddress: nodeBindAddress,
 		store:           store,
+		dialFunc:        driverDial,
 		driver:          driver,
 		driverName:      driverName,
 		log:             o.Log,
@@ -244,6 +260,19 @@ func New(dir string, options ...Option) (app *App, err error) {
 
 		cleanups = append(cleanups, func() { listener.Close(); <-proxyCh })
 
+	} else if o.Conn != nil {
+		go func() {
+			for {
+				remote := <-o.Conn.acceptCh
+				local, err := net.Dial("unix", nodeBindAddress)
+				if err != nil {
+					remote.Close()
+					panic(fmt.Errorf("Failed to connect to bind address %q: %w", nodeBindAddress, err))
+				}
+
+				go proxy(ctx, remote, local, nil)
+			}
+		}()
 	}
 
 	go app.run(ctx, o.RolesAdjustmentFrequency, joinFileExists)
@@ -647,11 +676,7 @@ func (a *App) makeRolesChanges(nodes []client.NodeInfo) RolesChanges {
 
 // Return the options to use for client.FindLeader() or client.New()
 func (a *App) clientOptions() []client.Option {
-	dial := client.DefaultDialFunc
-	if a.tls != nil {
-		dial = client.DialFuncWithTLS(dial, a.tls.Dial)
-	}
-	return []client.Option{client.WithDialFunc(dial), client.WithLogFunc(a.log)}
+	return []client.Option{client.WithDialFunc(a.dialFunc), client.WithLogFunc(a.log)}
 }
 
 func (a *App) debug(format string, args ...interface{}) {
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -1,6 +1,7 @@
 package app_test
 
 import (
+	"bufio"
 	"context"
 	"crypto/tls"
 	"crypto/x509"
@@ -8,6 +9,9 @@ import (
 	"encoding/binary"
 	"fmt"
 	"io/ioutil"
+	"net"
+	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 	"testing"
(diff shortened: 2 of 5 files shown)


MathieuBordere (commented):
LGTM, can you add some tests for the new option though?

MathieuBordere:
> LGTM, can you add some tests for the new option though? Something like creating a cluster with a couple of nodes and performing a write and a read on the db.

Thanks! Could you just add a 3rd node to the cluster in your test please? The reason is that the second node will remain in the cluster as a spare and the raft leader will not attempt to send it raft RPCs, thus not triggering `extDialFuncWithProxy` that is used by the raft nodes to talk to each other. Adding a 3rd node will trigger the non-leader nodes to be converted to voters which will cause `extDialFuncWithProxy` to be hit because the leader will start sending heartbeats to the other nodes.

So you should add 3rd node & wait a bit for the roles to settle.

masnax:
> > LGTM, can you add some tests for the new option though? Something like creating a cluster with a couple of nodes and performing a write and a read on the db.
> 
> Thanks! Could you just add a 3rd node to the cluster in your test please? The reason is that the second node will remain in the cluster as a spare and the raft leader will not attempt to send it raft RPCs, thus not triggering `extDialFuncWithProxy` that is used by the raft nodes to talk to each other. Adding a 3rd node will trigger the non-leader nodes to be converted to voters which will cause `extDialFuncWithProxy` to be hit because the leader will start sending heartbeats to the other nodes.
> 
> So you should add a 3rd node & wait a bit for the roles to settle.

Ah right, I forgot the spares don't do much :)
Added a third node and now they all come up as voters.

app/app_test.go
```diff
@@ -947,6 +951,85 @@ func newApp(t *testing.T, options ...app.Option) (*app.App, func()) {
...
+	require.NoError(t, err)
+	defer cli.Close()
+
+	cluster, err := cli.Cluster(context.Background())
+	require.NoError(t, err)
+	assert.Equal(t, externalAddr1, cluster[0].Address)
+	assert.Equal(t, externalAddr2, cluster[1].Address)
+	assert.Equal(t, externalAddr3, cluster[2].Address)
+
+	// Initially the node joins as spare.
```
MathieuBordere, line 1027: Comment is outdated.

MathieuBordere:
> > > LGTM, can you add some tests for the new option though? Something like creating a cluster with a couple of nodes and performing a write and a read on the db.
> > 
> > 
> > Thanks! Could you just add a 3rd node to the cluster in your test please? The reason is that the second node will remain in the cluster as a spare and the raft leader will not attempt to send it raft RPCs, thus not triggering `extDialFuncWithProxy` that is used by the raft nodes to talk to each other. Adding a 3rd node will trigger the non-leader nodes to be converted to voters which will cause `extDialFuncWithProxy` to be hit because the leader will start sending heartbeats to the other nodes.
> > So you should add a 3rd node & wait a bit for the roles to settle.
> 
> Ah right, I forgot the spares don't do much :) Added a third node and now they all come up as voters.

A minor nit about the comment and we should be good.

masnax:
> A minor nit about the comment and we should be good.

Fixed.

app/proxy.go
```diff
@@ -91,13 +98,47 @@ func proxy(ctx context.Context, remote net.Conn, local net.Conn, config *tls.Con
...
+
+	tcpConn, ok := c.(*net.TCPConn)
+	if !ok {
+		return nil, fmt.Errorf("Connection is not a net.TCPConn")
+	}
+
+	return tcpConn, nil
+}
+
 // Set TCP keepalive with 30 seconds idle time, 3 seconds retry interval with
```
tomponline, line 128: I think this comment needs updating.

app/proxy.go
```diff
@@ -117,6 +158,11 @@ func setKeepalive(conn *net.TCPConn) error {
 			if err != nil {
 				return
 			}
+
+			err = syscall.SetsockoptInt(fd, syscall.IPPROTO_TCP, unix.TCP_USER_TIMEOUT, int(30*time.Second/time.Millisecond))
```
tomponline, line 162: Adding a comment here detailing some of the reason behind its presence (from the comment in LXD's `SetTCPTimeouts`) would likely be beneficial for future readers.

app/app.go
```diff
@@ -244,6 +260,18 @@ func New(dir string, options ...Option) (app *App, err error) {
 
 		cleanups = append(cleanups, func() { listener.Close(); <-proxyCh })
 
+	} else if o.Conn != nil {
+		go func() {
+			for {
+				remote := <-o.Conn.acceptCh
+				local, err := net.Dial("unix", nodeBindAddress)
+				if err != nil {
+					continue
```
tomponline, line 269: Why are errors ignore here after accepting an inbound request?

app/dial.go
```diff
@@ -39,3 +39,21 @@ func makeNodeDialFunc(config *tls.Config) client.DialFunc {
 
 	return dial
 }
+
+func extDialFuncWithProxy(dialFunc client.DialFunc) client.DialFunc {
+	return func(ctx context.Context, addr string) (net.Conn, error) {
```
tomponline, line 44: Why does the function here accept a ctx, but doesn't pass it to the `go proxy()` function call inside it?

masnax: ah thanks I missed that.

app/app.go
tomponline: Just wondering also what we do with the accepted connection?

masnax: The remote accepted connection? We just copy any data back and forth between the accepted connection and dqlite's bind address. Did you mean something specific?

masnax: Yeah I don't see any problem with closing it.

app/dial.go
masnax: Hmm, actually it appears the parent context is already done by the time the dialfunc is run, which seems odd. Looking to find out why.

masnax: Ok so the parent context starts from the `connectWithDial` method in `internal/bindings/server.go`, which is called from `c`, and this is what executes the DialFunc. The context isn't cancelled until the DialFunc returns, but since we're running the proxy in a goroutine that we want to run as long as there's data between the connections,  we don't want the context to be cancelled by the caller.

tomponline: OK makes sense, so one is a connect timeout CTX and one is a cancelation context. 

What concerns me is why in one call to proxy() a cancelation context is passed as expected, but here the background context is passed meaning it won't get cancelled.

masnax: Ok, I added a context to the `Node` struct that starts/stops the dqlite server, which is then assigned to a variable in the `c` bindings when the server starts, and is cancelled when the server stops. 

This context becomes the parent context for the dial function, which in turn is only cancelled on a timeout or if the dial function returns any errors.

stgraber:
@masnax test seems to be getting stuck?

masnax:
> @masnax test seems to be getting stuck?

Yeah I just reversed the commit. Didn't like my change to contexts :(

tomponline:
> > @masnax test seems to be getting stuck?
> 
> Yeah I just reversed the commit. Didn't like my change to contexts :(

This got merged without a fix for the uncancelable context?

stgraber:
Gah, I missed that this is what got reverted, must have been checking an older set of commits against the new github action status. We should still wire up the ctx on the proxy to something other than a generic context.Background (but as mentioned, not pass the dial context through as that's not the right one for this).

masnax:
> > > @masnax test seems to be getting stuck?
> > 
> > 
> > Yeah I just reversed the commit. Didn't like my change to contexts :(
> 
> This got merged without a fix for the uncancelable context?

Yeah it did, at the very least it's not _more_ broken as the uncancelable context was already there, I just changed the lines above it :)

I dropped the last commit which added a life cycle context to the dqlite server as it was causing 1 test to hang, although I'm not exactly sure if the test was correct in the first place. 

@MathieuBordere Maybe you can help clarify,  `TestRolesAdjustment_CantReplaceVoter` in `app/app_test.go` seems to drop a 3-voter (and 1 stand-by) cluster to a 2-voter cluster, but I thought this should result in loss of quorum? I'm wondering if this test was only passing before because of the uncancelable context. Adding the voter back to the cluster will once again establish quorum and the test will pass.

MathieuBordere:
> > > > @masnax test seems to be getting stuck?
> > > 
> > > 
> > > Yeah I just reversed the commit. Didn't like my change to contexts :(
> > 
> > 
> > This got merged without a fix for the uncancelable context?
> 
> Yeah it did, at the very least it's not _more_ broken as the uncancelable context was already there, I just changed the lines above it :)
> 
> I dropped the last commit which added a life cycle context to the dqlite server as it was causing 1 test to hang, although I'm not exactly sure if the test was correct in the first place.
> 
> @MathieuBordere Maybe you can help clarify, `TestRolesAdjustment_CantReplaceVoter` in `app/app_test.go` seems to drop a 3-voter (and 1 stand-by) cluster to a 2-voter cluster, but I thought this should result in loss of quorum? I'm wondering if this test was only passing before because of the uncancelable context. Adding the voter back to the cluster will once again establish quorum and the test will pass.

So there are 3 voters and 1 standby node, then a voter and the standby go offline. The app logic will in this case not try to reshuffle the roles, the 2 remaining online nodes are voters and still have quorum (2 out of 3).
There's really not much else we can or should do, the standby was there to be promoted to voter in case a voter went offline, but in this test a voter and standby went offline at the same time.

PR #180, opened 2022-02-05
Don't use background context for c bindings.

masnax:
Adds a context to the dqlite server `Node` struct which starts on creation and is cancelled on close. This context is passed to the C bindings and stored in a `contextRegistry`, from which the `dialCtx` is derived.

As well, The `App` context is passed to the proxied dial functions for external connections and TLS.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -37,6 +37,7 @@ type App struct {
 	driver          *driver.Driver
 	driverName      string
 	log             client.LogFunc
+	ctx             context.Context
 	stop            context.CancelFunc // Signal App.run() to stop.
 	proxyCh         chan struct{}      // Waits for App.proxy() to return.
 	runCh           chan struct{}      // Waits for App.run() to return.
@@ -159,9 +160,10 @@ func New(dir string, options ...Option) (app *App, err error) {
 	}
 
 	// Start the local dqlite engine.
+	ctx, stop := context.WithCancel(context.Background())
 	var nodeDial client.DialFunc
 	if o.Conn != nil {
-		nodeDial = extDialFuncWithProxy(o.Conn.dialFunc)
+		nodeDial = extDialFuncWithProxy(ctx, o.Conn.dialFunc)
 	} else if o.TLS != nil {
 		nodeBindAddress = fmt.Sprintf("@dqlite-%d", info.ID)
 
@@ -173,7 +175,7 @@ func New(dir string, options ...Option) (app *App, err error) {
 			nodeBindAddress = fmt.Sprintf("@snap.%s.dqlite-%d", snapInstanceName, info.ID)
 		}
 
-		nodeDial = makeNodeDialFunc(o.TLS.Dial)
+		nodeDial = makeNodeDialFunc(ctx, o.TLS.Dial)
 	} else {
 		nodeBindAddress = info.Address
 		nodeDial = client.DefaultDialFunc
@@ -187,9 +189,11 @@ func New(dir string, options ...Option) (app *App, err error) {
 		dqlite.WithSnapshotParams(o.SnapshotParams),
 	)
 	if err != nil {
+		stop()
 		return nil, fmt.Errorf("create node: %w", err)
 	}
 	if err := node.Start(); err != nil {
+		stop()
 		return nil, fmt.Errorf("start node: %w", err)
 	}
 	cleanups = append(cleanups, func() { node.Close() })
@@ -204,6 +208,7 @@ func New(dir string, options ...Option) (app *App, err error) {
 
 	driver, err := driver.New(store, driver.WithDialFunc(driverDial), driver.WithLogFunc(o.Log))
 	if err != nil {
+		stop()
 		return nil, fmt.Errorf("create driver: %w", err)
 	}
 	driverIndex++
@@ -211,14 +216,14 @@ func New(dir string, options ...Option) (app *App, err error) {
 	sql.Register(driverName, driver)
 
 	if o.Voters < 3 || o.Voters%2 == 0 {
+		stop()
 		return nil, fmt.Errorf("invalid voters %d: must be an odd number greater than 1", o.Voters)
 	}
 
 	if o.StandBys%2 == 0 {
+		stop()
 		return nil, fmt.Errorf("invalid stand-bys %d: must be an odd number", o.StandBys)
 	}
-
-	ctx, stop := context.WithCancel(context.Background())
 
 	if runtime.GOOS != "linux" && nodeBindAddress[0] == '@' {
 		// Do not use abstract socket on other platforms and left trim "@"
@@ -237,6 +242,7 @@ func New(dir string, options ...Option) (app *App, err error) {
 		driverName:      driverName,
 		log:             o.Log,
 		tls:             o.TLS,
+		ctx:             ctx,
 		stop:            stop,
 		runCh:           make(chan struct{}, 0),
 		readyCh:         make(chan struct{}, 0),
@@ -270,7 +276,7 @@ func New(dir string, options ...Option) (app *App, err error) {
 					panic(fmt.Errorf("Failed to connect to bind address %q: %w", nodeBindAddress, err))
 				}
 
-				go proxy(ctx, remote, local, nil)
+				go proxy(app.ctx, remote, local, nil)
 			}
 		}()
 	}
@@ -451,7 +457,7 @@ func (a *App) Client(ctx context.Context) (*client.Client, error) {
 // Proxy incoming TLS connections.
 func (a *App) proxy() {
 	wg := sync.WaitGroup{}
-	ctx, cancel := context.WithCancel(context.Background())
+	ctx, cancel := context.WithCancel(a.ctx)
 	for {
 		client, err := a.listener.Accept()
 		if err != nil {
--- a/internal/bindings/server.go
+++ b/internal/bindings/server.go
@@ -85,7 +85,11 @@ import (
 	"github.com/canonical/go-dqlite/internal/protocol"
 )
 
-type Node C.dqlite_node
+type Node struct {
+	node   *C.dqlite_node
+	ctx    context.Context
+	cancel context.CancelFunc
+}
 
 type SnapshotParams struct {
 	Threshold uint64
@@ -113,7 +117,7 @@ func ConfigMultiThread() error {
 }
 
 // NewNode creates a new Node instance.
-func NewNode(id uint64, address string, dir string) (*Node, error) {
+func NewNode(ctx context.Context, id uint64, address string, dir string) (*Node, error) {
 	var server *C.dqlite_node
 	cid := C.dqlite_node_id(id)
 
@@ -128,15 +132,19 @@ func NewNode(id uint64, address string, dir string) (*Node, error) {
 		return nil, fmt.Errorf("%s", errmsg)
 	}
 
-	return (*Node)(unsafe.Pointer(server)), nil
+	node := &Node{node: (*C.dqlite_node)(unsafe.Pointer(server))}
+	node.ctx, node.cancel = context.WithCancel(ctx)
+
+	return node, nil
 }
 
 func (s *Node) SetDialFunc(dial protocol.DialFunc) error {
-	server := (*C.dqlite_node)(unsafe.Pointer(s))
+	server := (*C.dqlite_node)(unsafe.Pointer(s.node))
 	connectLock.Lock()
 	defer connectLock.Unlock()
 	connectIndex++
 	connectRegistry[connectIndex] = dial
+	contextRegistry[connectIndex] = s.ctx
 	if rc := C.configConnectFunc(server, connectIndex); rc != 0 {
 		return fmt.Errorf("failed to set connect func")
 	}
@@ -144,7 +152,7 @@ func (s *Node) SetDialFunc(dial protocol.DialFunc) error {
 }
 
 func (s *Node) SetBindAddress(address string) error {
-	server := (*C.dqlite_node)(unsafe.Pointer(s))
+	server := (*C.dqlite_node)(unsafe.Pointer(s.node))
 	caddress := C.CString(address)
 	defer C.free(unsafe.Pointer(caddress))
 	if rc := C.dqlite_node_set_bind_address(server, caddress); rc != 0 {
@@ -154,7 +162,7 @@ func (s *Node) SetBindAddress(address string) error {
 }
 
 func (s *Node) SetNetworkLatency(nanoseconds uint64) error {
-	server := (*C.dqlite_node)(unsafe.Pointer(s))
+	server := (*C.dqlite_node)(unsafe.Pointer(s.node))
 	cnanoseconds := C.nanoseconds_t(nanoseconds)
 	if rc := C.dqlite_node_set_network_latency(server, cnanoseconds); rc != 0 {
 		return fmt.Errorf("failed to set network latency")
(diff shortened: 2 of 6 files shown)


masnax:
One thing here has me scratching my head a bit:

`TestRolesAdjustment_CantReplaceVoter` in `app/app_test.go` seems to drop a 3-voter (and 1 stand-by) cluster to a 2-voter cluster, but I thought this should result in loss of quorum? I'm wondering if this test was only passing before because of the uncancelable context. Adding the voter back to the cluster will once again establish quorum and the test will pass.

internal/bindings/server.go
```diff
@@ -277,8 +278,17 @@ func connectWithDial(handle C.uintptr_t, address *C.char, fd *C.int) C.int {
 	defer connectLock.Unlock()
 	dial := connectRegistry[handle]
 	// TODO: make timeout customizable.
-	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
-	defer cancel()
+	ctx, cancel := context.WithTimeout(dialCtx, 5*time.Second)
+
+	// The dial function may be asynchronous so only cancel the context
+	// if we receive an error.
+	var err error
```
tomponline, line 285: Why not do this in the error handling of dial() below? Also I don't get your point about async here?

masnax: >Why not do this in the error handling of dial() below?
No reason, just thought to put all context related things in their own block.
>Also I don't get your point about async here?
Just that `dial` isn't guaranteed to be finished before its caller `connectWithDial`.

tomponline: How can that be? And how can you test the return value in that case?

Also the scoping on err may be problematic, especially on the call to connToSocket

masnax: So in the case of the TLS and external dial functions, (similar to LXD's `dqliteProxy`) when `dial` is run, it starts up a the `proxy` goroutine which in turn starts up two more goroutines that we expect to read and write from the connection that `dial` is returning until there's no more data.  So `dial` is returning a `net.Conn` that is still being used by a goroutine.

In practice, if we `defer cancel()` the parent context of the proxy inside of `connectWithDial`, the proxy will always end prematurely, before the `io.Copy` goroutines complete. 

>Also the scoping on err may be problematic, especially on the call to connToSocket

Not really a problem, since we would need to call a `cancel()` in the error block of `connToSocket` too.

tomponline: The err var being tested in the defer won't be the same one that is being returned from connToSocket, in that way it is a problem I think.

internal/bindings/server.go
```diff
@@ -291,6 +301,9 @@ func connectWithDial(handle C.uintptr_t, address *C.char, fd *C.int) C.int {
 	return C.int(0)
 }
 
+// Context of the parent node to be used by the dial function.
+var dialCtx context.Context
```
tomponline, line 305: This seems wrong to being a package level var in case there are multiple instances of the dqlite client.

tomponline:
I'll take a closer look at this when I next get on a PC as can't make much sense of it on the small phone screen 😀

app/app_test.go
```diff
@@ -726,19 +726,38 @@ func TestRolesAdjustment_CantReplaceVoter(t *testing.T) {
 	defer cleanups[0]()
 	defer cleanups[1]()
 
-	// A voter and a spare go offline.
```
MathieuBordere, line 729: As commented in https://github.com/canonical/go-dqlite/pull/179, this test tests that an offline voter's role is not transferred to another node if there is no other node to take it's role. This test should still succeed without adaptations.

internal/bindings/server.go
masnax: This was it! I didn't have a good grasp of how the server was starting up and how the dqlite C bindings talked to the rest of the Go code. The tests failed when a node went offline and I thought this was due to the cluster losing quorum but it was just this variable causing the wrong context to be cancelled.

internal/bindings/server.go
tomponline: OK so I've taken a closer look at this now that I'm back on my PC :)

My take from this is that in a nutshell we have two contexts at play:

1. The dial context - short lived (cancelled after 5s if not connected before)
2. The node context - long lived (cancelled when the node is stopped).

The thing I'm not quite clear on still though is that, when I call `func (s *Node) SetDialFunc(dial protocol.DialFunc, ctx context.Context)` am I providing the node context or the dial context?

Because on the one hand, its being stored in the `contextRegistry` suggesting a long-lived node context, but then `connectWithDial` is is used from the registry and named `dialCtx` (without extending it), and then has a 5s timeout added to it which is then called `ctx`. If the 5s timeout was the dialCtx, then I would have expected the naming inside `connectWithDial` to be the other way round (i.e the node ctx was loaded into `ctx` and then wrapped with a 5s timeout resulting in `dialCtx`).

I also don't understand why the `ctx, cancel := context.WithTimeout(dialCtx, 5*time.Second)` is cancelled on defer only if there has been an error. This is odd, as the context *will still* be cancelled after 5s even if there hasn't been an error (because it is `WithTimeout` type).

I think the friction here is that the formerly merged `extDialFuncWithProxy` function (https://github.com/canonical/go-dqlite/pull/179/files#diff-2b86d61eafcf5f028792eac153162dd06da4986de4d4c560d46cd51b85c75cf7R43) is returning a function that can only accept a single context (client.DialFunc) but really needs access to both; the dial context for the actual dial, and the long-lived node context for the long-lived proxy go routine that is started afterwards.

It feels to me like this could be solved by updating `makeNodeDialFunc` to accept the node's long-lived context from the `contextRegistry`, which could then be used by the `go proxy(ctx, conn, goUnix, clonedConfig)` line rather than using the dial context.

masnax: > The thing I'm not quite clear on still though is that, when I call func (s *Node) SetDialFunc(dial protocol.DialFunc, ctx context.Context) am I providing the node context or the dial context?

Poor naming there on my part, yes the `contextRegistry` holds stores the node context which is then used as the parent context for the dial context.

> It feels to me like this could be solved by updating makeNodeDialFunc to accept the node's long-lived context from the contextRegistry, which could then be used by the go proxy(ctx, conn, goUnix, clonedConfig) line rather than using the dial context.

I like this idea, though `makeNodeDialFunc` runs before there is any notion of a node so it would need to take the app context instead.

tomponline: That would probably be fine to have an app context rather than node context, as presumably if you remove a node that will cause its sockets be closed, causing its proxy go routine to end anyway right?

masnax:
Ok so as of now, the proxied TLS and external connection dial functions are using the context from `app.App`, and The actual dial function context `dialCtx` is derived from the `Node` context, which is stored in a `contextRegistry`.

internal/bindings/server.go
```diff
@@ -131,12 +131,13 @@ func NewNode(id uint64, address string, dir string) (*Node, error) {
 	return (*Node)(unsafe.Pointer(server)), nil
 }
 
-func (s *Node) SetDialFunc(dial protocol.DialFunc) error {
+func (s *Node) SetDialFunc(dial protocol.DialFunc, ctx context.Context) error {
```
tomponline, line 134: ctx should be first argument by convention.

internal/bindings/server.go
```diff
@@ -275,11 +276,15 @@ type fileConn interface {
 func connectWithDial(handle C.uintptr_t, address *C.char, fd *C.int) C.int {
 	connectLock.Lock()
 	defer connectLock.Unlock()
+
 	dial := connectRegistry[handle]
+	ctx := contextRegistry[handle]
+
 	// TODO: make timeout customizable.
-	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+	dialCtx, cancel := context.WithTimeout(ctx, 5*time.Second)
```
tomponline, line 284: As this function extends the node's ctx with a 5s timeout context, presumably, for each dial attempt, then we can say that the node's ctx is a long-lived one. But, now that we are using the app's context to cancel the node's proxy go routine, can/should we also use the app's context to cancel (prematurely) the node's dial attempt also (and avoid the need for a per-node ctx).

Basically what I'm aiming for is consistency. If we have a long-live node context, then it should be used as a basis for all node related operations (in this case the dial attempt and the long-lived proxy routine).

But if we are now using the app's context for the long-lived node proxy routine, then for consistency/ease of understanding, we should also use the app's context for the basis of the 5s dial timeout context too.

The question is, I think, is a node stopped (which can cancel its node ctx) when the app continues running, and in those situations, should we have cancelled any ongoing dial operations as well as any long-live proxy routines?

masnax: I left the contexts separate because we can create, start up and close a dqlite node without using the app package at all, like we do in LXD. Though in a way, LXD is taking the role of the app package in that case, so perhaps we can add a `WithContext` option to the options passed to `dqlite.New` that overrides the background context as the parent context for the dqlite node. Then if it makes sense for LXD, we can use `WithContext` to set the gateway context to the parent context of the dqlite node.

Of course, we can force consideration of the context of dqlite by just adding a context argument to `dqlite.New`, but this would mean we would have to update LXD to use this before we can update the dqlite version.

tomponline: I'm fine with just having a node context, but we should then ensure the proxy() function is passed it rather than the app context as it does now.

masnax: Since the dial functions were being configured before the dqlite server was starting, I moved the logic for configuring the dial funcs and starting up the proxy into the dqlite package itself. So now the context passed to the dial function (the one used for the proxies) has the node context as its parent. 

I moved the node.go file into its own directory just to make things a bit cleaner (as now there's 4 files in there), but this changes the path that dqlite is currently imported.

masnax: Sorry, I didn't realize the diff would be so horrible. I've moved the dqlite package stuff back out into the root of the project.

Bit confused about your suggestion for `SetDialFunc`. There's unfortunately 2 `Node` types which makes things a bit confusing.  I think you mean adding context fields to the `Node` struct for the bindings, as that's the owner of `SetDialFunc`. This works, but I think the `contextRegistry` would still be necessary since `connectWithDial` will eventually need to call the dial function with the right context and it's not a method on `Node`.

masnax: What do you think about creating the context in `Start` instead? I'm not so sure about starting a 5 second context in `New`, which will be counting down right away, even though we're not even dialing yet.

tomponline: We could have it in Start too I guess, but New seems better as it will be initialised even if Start isn't called.

masnax: So now,`dqlite.Node` creates a new context from `context.Background()` in `New`, and passes it to `server.NewNode`, which stores it in the context registry.

internal/bindings/server.go
```diff
@@ -135,7 +140,8 @@ func (s *Node) SetDialFunc(dial protocol.DialFunc) error {
 	server := (*C.dqlite_node)(unsafe.Pointer(s))
 	connectLock.Lock()
 	defer connectLock.Unlock()
-	connectIndex++
+
+	// The index should have been incremented by NewNode.
```
MathieuBordere, line 144: This will not work, what happens when someone does? (pseudocode)

```
n1 := NewNode()
n2 := NewNode()
n1.SetDialFunc(func)
n2.SetDialFunc(func)
```

app/app.go
```diff
@@ -173,22 +169,35 @@ func New(dir string, options ...Option) (app *App, err error) {
...
 		dqlite.WithNetworkLatency(o.NetworkLatency),
 		dqlite.WithSnapshotParams(o.SnapshotParams),
 	)
 	if err != nil {
 		return nil, fmt.Errorf("create node: %w", err)
 	}
+
+	// Set the appropriate dqlite dial function before starting.
+	if o.Conn != nil {
+		err = node.SetExternalDialFunc(o.Conn.dialFunc)
```
MathieuBordere, line 190: Was there no way to keep using `WithDialFunc` ?

internal/bindings/server.go
tomponline: I think I may be confusing the situation, when I get a chance I'll take a crack at this and hopefully will be able to provide better explanations :)

internal/bindings/server.go
masnax: Wishful thinking :)

Changed it to just assign the context to the `Node` struct now.

app/app.go
masnax: We want both the Node context (parent) and 5-second timed dial context (child) to be exposed to the dial function. We can't pass in the former if the dial func has already been defined, so either we can't use `WithDialFunc` or we would have to add a `parentCtx` parameter to `client.DialFunc`.

tomponline:
@masnax I'll take a look at this later today or early next week. Sorry for the delay.

tomponline:
@masnax I got a chance to (briefly) look at this just now and its indeed very confusing given the way the C code is interacting with the Go code and vice versa.

One thing that caught my eye though is the `proxy()` function. Is that analogous to the `dqliteProxy()` function in LXD (it certainly looks like it)? If so then I don't think we actually need the `ctx` argument on `proxy()` at all (similarly to how the `dqliteProxy()` function does have one).

The reason for this is that there are already two ways to stop that function, by closing either of `remote` or `local` connections, which in turn closes the other one.

And if neither of the connections have been closed and the app is shutting down, then there is a problem elsewhere in the app/package because closing the client shouldn't be handled in the proxy anyway AFAIK.

If we can remove the `ctx` argument from that `proxy` function then the rest of the PR goes away :)

As for ignoring the errors from `proxy` when the ctx is cancelled, I think we can more cleanly handle that by detecting of `ctx.Error() != nil` when `proxy` returns an error, and if so ignore it, because by then you should have access to the App ctx (which may need to be stored if not already).

What do you think?

masnax:
> @masnax I got a chance to (briefly) look at this just now and its indeed very confusing given the way the C code is interacting with the Go code and vice versa.
> 
> One thing that caught my eye though is the `proxy()` function. Is that analogous to the `dqliteProxy()` function in LXD (it certainly looks like it)? If so then I don't think we actually need the `ctx` argument on `proxy()` at all (similarly to how the `dqliteProxy()` function does have one).
> 
> The reason for this is that there are already two ways to stop that function, by closing either of `remote` or `local` connections, which in turn closes the other one.
> 
> And if neither of the connections have been closed and the app is shutting down, then there is a problem elsewhere in the app/package because closing the client shouldn't be handled in the proxy anyway AFAIK.
> 
> If we can remove the `ctx` argument from that `proxy` function then the rest of the PR goes away :)
> 
> As for ignoring the errors from `proxy` when the ctx is cancelled, I think we can more cleanly handle that by detecting of `ctx.Error() != nil` when `proxy` returns an error, and if so ignore it, because by then you should have access to the App ctx (which may need to be stored if not already).
> 
> What do you think?

Seems reasonable to get rid of the proxy context. However, if we get rid of the entire PR, then the context passed to `dial` in `connectWithDial` will still have a background context as its parent. Do you think it's worthwhile to leave in the `contextRegistry` bits so that the Node context is still the parent of the dial Context?

tomponline:
Yes that makes sense to have the dial context passable.

masnax:
@stgraber Sorry, jammed the tests :(

masnax:
So now the `proxy` method is using a `stopCh` in lieu of the context. Since the conns the proxy is using are both created as part of the dial func, we can't just have the app package close them. Instead the `stopCh` is telling the proxy to do it for us. This method now looks exactly like its lxd counterpart `dqliteProxy` :)

masnax:
> At this risk of seemingly going round in circles.

Yup, actually passing the app context into the helpers that create the dial func was something we tried earlier :)

tomponline:
> > At this risk of seemingly going round in circles.
> 
> Yup, actually passing the app context into the helpers that create the dial func was something we tried earlier :)
 
Maybe I couldn't see the woods from the trees as was in the new package, this smaller pr is much easier to reason about.

tomponline (approved):
Cool thanks for persevering with me!

PR #181, opened 2022-02-10
External Connection: Write hijacked response header

masnax:
Signed-off-by: Max Asnaashari <max.asnaashari@canonical.com>


stgraber:
Looks good to me. This would likely break existing users if it wasn't for a bit of code that's just a few days old and that never made it in a release yet, so I think that's fine.

PR #184, opened 2022-02-25
app: Add WithUnixSocket

stgraber:
Signed-off-by: Stéphane Graber <stephane.graber@canonical.com>


stgraber:
This is needed because when running a snap, the set of addresses we can bind is limited, so a random abstract unix socket won't work.

stgraber:
Confirmed that this works perfectly for the lxd-cloud snap.

@tomponline @MathieuBordere good to go

PR #185, opened 2022-03-16
app: Export SnapshotParams

MathieuBordere:


node.go
```diff
@@ -23,6 +23,9 @@ type Node struct {
 // NodeInfo is a convenience alias for client.NodeInfo.
 type NodeInfo = client.NodeInfo
 
+// Expose SnapshotParams.
```
stgraber, line 26: If exported, it can probably do with a more useful docstring :)

Issue #182, opened 2022-02-10
Outbound connection does not appear to be closed/removed from use when remote side closes it


tomponline:
See https://github.com/lxc/lxd/pull/9891 for more info.

The reproducer is fairly easy:

1. Launch 3 Ubuntu Focal VMs and install the lxd edge snap (sudo snap install lxd --edge).
2. Setup a 3 node LXD cluster (sudo lxd init).
3. Once cluster is up and running, shutdown the leader member cleanly (i.e lxc stop and not lxc stop -f).
4. On the the third member (the one not promoted to leader), check sudo ss -tnp and you should see an outbound LXD 8443 connection to the shutdown member in close-wait state (the logging this PR adds has identified that it is an outbound dqlite and not raft connection).
5. Wait a few seconds and then try running lxc cluster ls on the non-leader member, it should start giving the error at that point.

```
Error: failed to begin transaction: call exec-sql (budget 9.999963394s): receive: header: EOF
```

This doesn't happen 100% of the time, but trying a couple of times should reproduce it, making sure you always test on a non-leader member.

This issue has appeared since we removed the continuous background job for maintaining the LXD event websocket connections (which was causing every member to query the cluster database for members every 1s), somehow by stopping those continuous queries, this has prevented go-dqlite from detecting/handling a remotely closed connection.

My hunch is that its an issue in driverError in go-dqlite, somehow not handling/detecting the EOF error and then not returning driver.ErrBadConn err that would presumably cause the connection to be taken out of use by dqlite.

MathieuBordere:
Possibly related to be implemented in conjunction with this change.

https://pkg.go.dev/database/sql/driver

> All Conn implementations should implement the following interfaces: Pinger, SessionResetter, and Validator. 

```
type SessionResetter interface {
	// ResetSession is called prior to executing a query on the connection
	// if the connection has been used before. If the driver returns ErrBadConn
	// the connection is discarded.
	ResetSession(ctx [context](https://pkg.go.dev/context).[Context](https://pkg.go.dev/context#Context)) [error](https://pkg.go.dev/builtin#error)
}
```
```
type Validator interface {
	// IsValid is called prior to placing the connection into the
	// connection pool. The connection will be discarded if false is returned.
	IsValid() [bool](https://pkg.go.dev/builtin#bool)
}
```
```

PR #186, opened 2022-04-26
driver: Detect EOF in driverError

MathieuBordere:
Fixes https://github.com/canonical/go-dqlite/issues/182

### DIFF
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -787,6 +787,12 @@ type unwrappable interface {
 	Unwrap() error
 }
 
+// TODO driver.ErrBadConn should not be returned when there's a possibility that
+// the query has been executed. In our case there is a window in protocol.Call
+// between `send` and `recv` where the send has succeeded but the recv has
+// failed. In those cases we call driverError on the result of protocol.Call,
+// possibly returning ErrBadCon.
+// https://cs.opensource.google/go/go/+/refs/tags/go1.20.4:src/database/sql/driver/driver.go;drc=a32a592c8c14927c20ac42808e1fb2e55b2e9470;l=162
 func driverError(log client.LogFunc, err error) error {
 	switch err := errors.Cause(err).(type) {
 	case syscall.Errno:
@@ -836,6 +842,10 @@ func driverError(log client.LogFunc, err error) error {
 			log(client.LogDebug, "network connection lost: %v", err)
 			return driver.ErrBadConn
 		}
+	}
+	if errors.Is(err, io.EOF) {
+		log(client.LogDebug, "EOF detected: %v", err)
+		return driver.ErrBadConn
 	}
 	return err
 }
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -353,7 +353,7 @@ func (c *Conn) PrepareContext(ctx context.Context, query string) (driver.Stmt, e
 	if c.tracing != client.LogNone {
 		start = time.Now()
 	}
-	err := c.protocol.Call(ctx, &c.request, &c.response);
+	err := c.protocol.Call(ctx, &c.request, &c.response)
 	if c.tracing != client.LogNone {
 		c.log(c.tracing, "%.3fs request prepared: %q", time.Since(start).Seconds(), query)
 	}
@@ -392,7 +392,7 @@ func (c *Conn) ExecContext(ctx context.Context, query string, args []driver.Name
 	if c.tracing != client.LogNone {
 		start = time.Now()
 	}
-	err := c.protocol.Call(ctx, &c.request, &c.response);
+	err := c.protocol.Call(ctx, &c.request, &c.response)
 	if c.tracing != client.LogNone {
 		c.log(c.tracing, "%.3fs request exec: %q", time.Since(start).Seconds(), query)
 	}
@@ -428,7 +428,7 @@ func (c *Conn) QueryContext(ctx context.Context, query string, args []driver.Nam
 	if c.tracing != client.LogNone {
 		start = time.Now()
 	}
-	err := c.protocol.Call(ctx, &c.request, &c.response);
+	err := c.protocol.Call(ctx, &c.request, &c.response)
 	if c.tracing != client.LogNone {
 		c.log(c.tracing, "%.3fs request query: %q", time.Since(start).Seconds(), query)
 	}
@@ -588,7 +588,7 @@ func (s *Stmt) ExecContext(ctx context.Context, args []driver.NamedValue) (drive
 	if s.tracing != client.LogNone {
 		start = time.Now()
 	}
-	err := s.protocol.Call(ctx, s.request, s.response);
+	err := s.protocol.Call(ctx, s.request, s.response)
 	if s.tracing != client.LogNone {
 		s.log(s.tracing, "%.3fs request prepared: %q", time.Since(start).Seconds(), s.sql)
 	}
@@ -627,7 +627,7 @@ func (s *Stmt) QueryContext(ctx context.Context, args []driver.NamedValue) (driv
 	if s.tracing != client.LogNone {
 		start = time.Now()
 	}
-	err := s.protocol.Call(ctx, s.request, s.response);
+	err := s.protocol.Call(ctx, s.request, s.response)
 	if s.tracing != client.LogNone {
 		s.log(s.tracing, "%.3fs request prepared: %q", time.Since(start).Seconds(), s.sql)
 	}


tomponline:
@MathieuBordere once this is merged, please can you create a new release version and tag for this so we can update LXD's dependencies ready for LXD 5.1?

freeekanayaka:
With this change you would be able to re-merge the dqlite change that was reverted, right?

tomponline:
> With this change you would be able to re-merge the dqlite change that was reverted, right?

Yes indeed.

MathieuBordere:
But maybe I should test for EOF at the end of the other error handling? I'm just a bit confused by `err` being reassigned a couple of times e.g. do we get the original err back here? https://github.com/canonical/go-dqlite/blob/5724311830ff770590d89ae48b2ea40c6566dc95/driver/driver.go#L775

tomponline:
I think that it is fine as it is as it will find EOF anywhere in the error chain.

tomponline:
@MathieuBordere can you fix up the test golint issues please (looks like just comment annotation issues)

freeekanayaka (approved):
Looks good to me.

It somehow reinforces the feeling that in the future we should also work on proper error propagation vs closing connections, but I'm happy that it solves the problem for now.

MathieuBordere:
> @MathieuBordere can you fix up the test golint issues please (looks like just comment annotation issues)
> 
> https://github.com/canonical/go-dqlite/pull/186/files "Unchanged files with check annotations Beta "

Done, the `expected type, found '='` in `type NodeInfo = client.NodeInfo` doesn't show up locally running golint and is maybe related to https://github.com/golang/lint/issues/335

freeekanayaka:
> But maybe I should test for EOF at the end of the other error handling? I'm just a bit confused by `err` being reassigned a couple of times e.g. do we get the original `err` back here?
> 
> https://github.com/canonical/go-dqlite/blob/5724311830ff770590d89ae48b2ea40c6566dc95/driver/driver.go#L775

You have a good point. Yes, I do think it's a good idea to place this at the end of the chain, in the default statement, and yes you should be fine with the `err` object returned by `root.Unwrap()`.

So hopefully something like:

```Go
default:
		// When using a TLS connection, the underlying error might get
		// wrapped by the stdlib itself with the new errors wrapping
		// conventions available since go 1.13. In that case we check
		// the underlying error with Unwrap() instead of Cause().
		if root, ok := err.(unwrappable); ok {
			err = root.Unwrap()
		}
		switch err.(type) {
		case *net.OpError:
                        fallthrough
		case *io.EOF:
			log(client.LogDebug, "network connection lost: %v", err)
			return driver.ErrBadConn
		}
```

should work.

freeekanayaka:
One thing that concerns me is that I believe the proxy (both the LXD one and the `go-dqlite/app` one) might actually depend on getting back an `io.EOF` for graceful stopping and cleanup. It might be just a speculation, but the whole mechanism seems to have gotten a bit wild and might need some attention, designing and implementing a good system for leadership and network-related error handling and propagation, as we were saying in the other PR too.

tomponline:
I suppose if the lxd tests pass OK then it should be fine

MathieuBordere:
> I suppose if the lxd tests pass OK then it should be fine

suite is running now, will report back.

tomponline:
There are also tests in this repo that check that the result set iterators return io.EOF at the end of the rows, so if they pass that will be encouraging. But moving the check to the `default:` section shouldn't hurt either if it makes the more specific error handlers have an opportunity to check first.

MathieuBordere:
I will mark this as draft for now, and investigate further before quickly patching up the issue uncovered by that dqlite PR, don't want to add to the mess.

I will handle the dqlite leadership loss in a different way than closing the connection.

tomponline:
> I will handle the dqlite leadership loss in a different way than closing the connection.

Which way are you thinking?

MathieuBordere:
> > I will handle the dqlite leadership loss in a different way than closing the connection.
> 
> Which way are you thinking?

It will close the SQLite leader connection to the database, eliminating the `Assert` you see in the lxd testsuite, it will finish the ongoing SQL request with a `LEADERSHIP_LOST` and all future db queries on that connection will return `LEADERSHIP_LOST` causing the driver to mark the connection as bad see https://github.com/canonical/go-dqlite/blob/5724311830ff770590d89ae48b2ea40c6566dc95/driver/driver.go#L752.

freeekanayaka:
> > > I will handle the dqlite leadership loss in a different way than closing the connection.
> > 
> > 
> > Which way are you thinking?
> 
> It will close the SQLite leader connection to the database, eliminating the `Assert` you see in the lxd testsuite, it will finish the ongoing SQL request with a `LEADERSHIP_LOST` and all future db queries on that connection will return `LEADERSHIP_LOST` causing the driver to mark the connection as bad see
> 
> https://github.com/canonical/go-dqlite/blob/5724311830ff770590d89ae48b2ea40c6566dc95/driver/driver.go#L752

Looking again at the description of  #354, I feel there might be a bit of confusion about what `LEADERSHIP_LOST` is supposed to mean/do.

Essentially, `LEADERSHIP_LOST` should be returned in case the client asks for a commit and that commit can't be executed because the node lost leadership while waiting for a quorum the frames command log entry that it had produced as soon as it received the commit request from the client.

In that case the client can't make any assumption about whether the commit was (or will) actually successful, because there's no way to know if a quorum of nodes had received the entry. This is basically the situation described in section 6.3 (Implementing linearizable semantics) of the Raft dissertation, and the solution indicated there would be to implement client sessions and request IDs in the FSM. With that in place, the client would retry the commit request as soon as it finds a new leader, using sessions and request IDs the new leader would know if it is a brand new commit request, or actually a duplicated one (since the original one got committed despite the client received a `LEADERSHIP_LOST` error). This design is also described in the [dqlite docs](https://dqlite.io/docs/replication) (see the "Client sessions" section at the end of the page), but it is still to be implemented.

Now, after having described what `LEADERSHIP_LOST` was meant to be, let's look at the problematic sequence of events described in #354 :

* Node 1 is leader, Node 2 is follower
* Node 1 receives BEGIN - QUERY XXXX - ... (notice that it doesn't receive a COMMIT message)
* Node 1 transfers leadership to Node 2, or loses leadership in some way

^^^ at this point what I believe should happen is:

* For the client: as soon as it tries to send the COMMIT message it will get back a `NOT_LEADER` error, since the gateway will notice that it's not the leader anymore. The Go driver should already handle that and propagate a `driver.ErrBadConn` so the connection gets closed by the client itself (both the network connection and the associated internal SQLite object/connection) with no need to do anything special from the server. 
* For the server: as soon as it detects that the node is not anymore the leader (using the `monitor_cb`) it rolls back any uncommitted transactions in the VFS, using `dqlite_vfs_abort`.

* Node 2 writes some data to the db
* Node 1 receives the frames, from Node 2, tries to apply them to the DB and triggers an Assert due to the pending transaction in the leader connection from step 2 that is holding locks. 

^^^ at this point the assertion shouldn't be triggered since the server had rolled back that pending transaction.

What I described should result in relatively small changes in the `monitor_cb` to issue the relevant `dqlite_vfs_abort` calls, and nothing else. In the future we might want to tackle the much larger problem of proper `LEADERSHIP_LOST` handler using client sessions and request IDs.

MathieuBordere:
> > > > I will handle the dqlite leadership loss in a different way than closing the connection.
> > > 
> > > 
> > > Which way are you thinking?
> > 
> > 
> > It will close the SQLite leader connection to the database, eliminating the `Assert` you see in the lxd testsuite, it will finish the ongoing SQL request with a `LEADERSHIP_LOST` and all future db queries on that connection will return `LEADERSHIP_LOST` causing the driver to mark the connection as bad see
> > https://github.com/canonical/go-dqlite/blob/5724311830ff770590d89ae48b2ea40c6566dc95/driver/driver.go#L752
> 
> Looking again at the description of #354, I feel there might be a bit of confusion about what `LEADERSHIP_LOST` is supposed to mean/do.
> 
> Essentially, `LEADERSHIP_LOST` should be returned in case the client asks for a commit and that commit can't be executed because the node lost leadership while waiting for a quorum of the frames command log entry that it had produced as soon as it received the commit request from the client.
> 
> In that case the client can't make any assumption about whether the commit was (or will be) actually successful, because there's no way to know if a quorum of nodes had received the entry. This is basically the situation described in section 6.3 (Implementing linearizable semantics) of the Raft dissertation, and the solution indicated there would be to implement client sessions and request IDs in the FSM. With that in place, the client would retry the commit request as soon as it finds a new leader, using sessions and request IDs the new leader would know if it is a brand new commit request, or actually a duplicated one (since the original one got committed despite the client received a `LEADERSHIP_LOST` error). This design is also described in the [dqlite docs](https://dqlite.io/docs/replication) (see the "Client sessions" section at the end of the page), but it is still to be implemented.
> 
> Now, after having described what `LEADERSHIP_LOST` was meant to be, let's look at the problematic sequence of events described in #354 :
> 
>     * Node 1 is leader, Node 2 is follower
> 
>     * Node 1 receives BEGIN - QUERY XXXX - ... (notice that it doesn't receive a COMMIT message)
> 
>     * Node 1 transfers leadership to Node 2, or loses leadership in some way
> 
> 
> ^^^ at this point what I believe should happen is:
> 
>     * For the client: as soon as it tries to send the COMMIT message it will get back a `NOT_LEADER` error, since the gateway will notice that it's not the leader anymore. The Go driver should already handle that and propagate a `driver.ErrBadConn` so the connection gets closed by the client itself (both the network connection and the associated internal SQLite object/connection) with no need to do anything special from the server.
> 
>     * For the server: as soon as it detects that the node is not anymore the leader (using the `monitor_cb`) it rolls back any uncommitted transactions in the VFS, using `dqlite_vfs_abort`.
> 
> 
> And then, proceeding in the problematic sequence of events:
> 
>     * Node 2 writes some data to the db
> 
>     * Node 1 receives the frames, from Node 2, tries to apply them to the DB and triggers an Assert due to the pending transaction in the leader connection from step 2 that is holding locks.
> 
> 
> ^^^ at this point the assertion shouldn't be triggered since the server had rolled back that pending transaction.
> 
> What I described should result in relatively small changes in the `monitor_cb` to issue the relevant `dqlite_vfs_abort` calls, and nothing else. In the future we might want to tackle the much larger problem of proper `LEADERSHIP_LOST` handling using client sessions and request IDs.

I understand your reasoning but `VfsAbort` doesn't get rid off the Assert, it's a read lock that is still being held in the test, it's just a read transaction.

`integration-test: src/vfs.c:2372: vfsInvalidateWalIndexHeader: Assertion `shm->shared[4] == 0' failed.`

while `VfsAbort` basically just releases an exclusive lock 
https://github.com/canonical/dqlite/blob/bc1cb2d51cb7e0c3c72d3233c3dfa5c1453d33a8/src/vfs.c#L2449

I thought that closing the sqlite connection would be a fool proof way to get rid off all the state introduced by the start of the transaction.

freeekanayaka:
> I understand your reasoning but `VfsAbort` doesn't get rid off the Assert, it's a read lock that is still being held in the test, it's just a read transaction.
> 
> `integration-test: src/vfs.c:2372: vfsInvalidateWalIndexHeader: Assertion 'shm->shared[4] == 0' failed.`
> 
> while `VfsAbort` basically just releases an exclusive lock https://github.com/canonical/dqlite/blob/bc1cb2d51cb7e0c3c72d3233c3dfa5c1453d33a8/src/vfs.c#L2449
> 
> I thought that closing the sqlite connection would be a fool proof way to get rid off all the state introduced by the start of the transaction.

Yeah, I was going to follow-up since I just realized that. Indeed `VfsAbort` should be used in the future for `LEADERSHIP_LOST` error, because in that case there will be a write lock.

In this case, instead of using `VfsAbort`, I'd suggest using a normal "ROLLBACK" SQL statement, e.g. `sqlite3_exec(conn, "ROLLBACK", NULL, NULL, NULL).` That should do it.

freeekanayaka:
Closing the connection altogether, probably works too, but we'd end up in bit of a mixed state in the gateway, which feels potentially a bit fragile, or will always need some care (i.e. we have to know that there are cases where the leader object or SQLite object is not there because it was closed elsewhere). Having a single path for releasing resource usually simplifies things.

MathieuBordere:
> > I understand your reasoning but `VfsAbort` doesn't get rid off the Assert, it's a read lock that is still being held in the test, it's just a read transaction.
> > `integration-test: src/vfs.c:2372: vfsInvalidateWalIndexHeader: Assertion 'shm->shared[4] == 0' failed.`
> > while `VfsAbort` basically just releases an exclusive lock https://github.com/canonical/dqlite/blob/bc1cb2d51cb7e0c3c72d3233c3dfa5c1453d33a8/src/vfs.c#L2449
> > I thought that closing the sqlite connection would be a fool proof way to get rid off all the state introduced by the start of the transaction.
> 
> Yeah, I was going to follow-up since I just realized that. Indeed `VfsAbort` should be used in the future for `LEADERSHIP_LOST` error, because in that case there will be a write lock.
> 
> In this case, instead of using `VfsAbort`, I'd suggest using a normal "ROLLBACK" SQL statement, e.g. `sqlite3_exec(conn, "ROLLBACK", NULL, NULL, NULL).` That should do it.

Yeah, I thought about this too, but there is a case where this can lead to inconsistent data I think.

- Client issues BEGIN
- Client issues some queries
- Client loses leadership
- We abort the transaction with ROLLBACK
- SQLite is now in autocommit mode
- Client regains leadership
- Client continues queries (these will autocommit, while the queries from earlier will not be comitted)
- Client issues COMMIT (this wil fail, as there is no BEGIN).

freeekanayaka:
> > > I understand your reasoning but `VfsAbort` doesn't get rid off the Assert, it's a read lock that is still being held in the test, it's just a read transaction.
> > > `integration-test: src/vfs.c:2372: vfsInvalidateWalIndexHeader: Assertion 'shm->shared[4] == 0' failed.`
> > > while `VfsAbort` basically just releases an exclusive lock https://github.com/canonical/dqlite/blob/bc1cb2d51cb7e0c3c72d3233c3dfa5c1453d33a8/src/vfs.c#L2449
> > > I thought that closing the sqlite connection would be a fool proof way to get rid off all the state introduced by the start of the transaction.
> > 
> > 
> > Yeah, I was going to follow-up since I just realized that. Indeed `VfsAbort` should be used in the future for `LEADERSHIP_LOST` error, because in that case there will be a write lock.
> > In this case, instead of using `VfsAbort`, I'd suggest using a normal "ROLLBACK" SQL statement, e.g. `sqlite3_exec(conn, "ROLLBACK", NULL, NULL, NULL).` That should do it.
> 
> Yeah, I thought about this too, but there is a case where this can lead to inconsistent data I think.
> 
>     * Client issues BEGIN
> 
>     * Client issues some queries
> 
>     * Node to which client is connected loses leadership
> 
>     * We abort the transaction with ROLLBACK
> 
>     * SQLite is now in autocommit mode
> 
>     * Node to which client is connected regains leadership
> 
>     * Client continues queries (these will autocommit, while the queries from earlier will not be comitted)
> 
>     * Client issues COMMIT (this wil fail, as there is no BEGIN).

Okay, thanks for pointing this out. I actually had also another idea, which I felt was actually better but a bit more subtle. Your point makes me feel that perhaps it's actually the best approach.

Basically we would not do anything in the `monitor_cb` when the raft state changes, that's because there's always the chance that the node regains leadership and everything can proceed just as it was (like in your example).

Instead we should turn the `assert(shm->shared[i] == 0)` in `vfsInvalidateWalIndexHeader()` into a check: "if there are pending read transactions, rollback them, assuming they were started when this node was the leader and we do know have prove that there is another leader that has sent us new frames".

The only two things that I'm not totally sure about are:

1. Do we have a way to turn that "assuming they were started when this node was the leader" part into something we can actually check instead of assuming? To avoid masking genuine cases where the assertion is valid and we should blow up because the read transaction was not started when the node was a leader.
2. What does it mean to rollback here, probably releasing the read lock using `vfsShmUnlock()` should be enough, but I didn't check.

Point 1. is kind of minor, but something that would be nice to have.

MathieuBordere:
> I don't really feel too comfortable with that, it feels a bit brittle as I don't fully understand it, I'd rather have SQLite itself do the cleanups for us for whatever state it introduced. I think that a case can be made for the fact that a connection to a leader node that at one point has lost leadership is "busted", shouldn't be reused and cleaned up.
> 
> I would propose to close the sqlite leader connection and add extra checks in functions that now assume that `gateway->leader` can never be NULL.

freeekanayaka:
If you close the SQLite connection only (and I guess destroy the `struct leader` object too?), and not the network one, how do you communicate that to the client? Assuming that no information gets sent to the client within the context of the `monitor_cb` (but correct me if that's not what you think), then the client will eventually try to issue a `COMMIT` statement, and what happens in that case if the node has regained leadership in the meantime? (as per your example).

MathieuBordere:
> If you close the SQLite connection only (and I guess destroy the `struct leader` object too?), and not the network one, how do you communicate that to the client? Assuming that no information gets sent to the client within the context of the `monitor_cb` (but correct me if that's not what you think), then the client will eventually try to issue a `COMMIT` statement, and what happens in that case if the node has regained leadership in the meantime? (as per your example).

I would always reply "LEADERSHIP_LOST" in that case (or some other error message if you prefer), basically the connection will be unusable and should be interpreted that way by the client/driver.

MathieuBordere:
> > > If you close the SQLite connection only (and I guess destroy the `struct leader` object too?), and not the network one, how do you communicate that to the client? Assuming that no information gets sent to the client within the context of the `monitor_cb` (but correct me if that's not what you think), then the client will eventually try to issue a `COMMIT` statement, and what happens in that case if the node has regained leadership in the meantime? (as per your example).
> > 
> > 
> > I would always reply "LEADERSHIP_LOST" in that case (or some other error message if you prefer), basically the connection will be unusable and should be interpreted that way by the client/driver.
> 
> Okay, so essentially you would set a flag on the `gateway` object (like the `leadership_lost` flag in your very first PR)? I would suggest then to not use `LEADERSHIP_LOST` (since it has a different semantics, i.e. "retry with new leader"), but instead use something else, e.g. `SQLITE_NOTFOUND` or anything more appropriate (e.g. one of the `SQLITE_IOERR_XXX` codes).
> 
> Or maybe no flag is needed, since the `LOOKUP_DB(ID)` call in `gateway.c` already checks for `req->gateway->leader == NULL` and returns `SQLITE_NOTFOUND` in that case.

Ok, I'll open a PR for that.

MathieuBordere:
Specifically, will check if the change (and existing code) violates the rules for ErrBadCon layed out in the [Docs](https://pkg.go.dev/database/sql)

> ErrBadConn should only be returned from Validator, SessionResetter, or a query method if the connection is already in an invalid (e.g. closed) state.
> ErrBadConn should be returned by a driver to signal to the sql package that a driver.Conn is in a bad state (such as the server having earlier closed the connection) and the sql package should retry on a new connection.
> To prevent duplicate operations, ErrBadConn should NOT be returned if there's a possibility that the database server might have performed the operation. Even if the server sends back an error, you shouldn't return ErrBadConn.

freeekanayaka:
Please would you recap what the issue exactly is? Like what conditions lead to it. A reproducer or a unit test would be great too, although I guess that might be tricky.

 I've tried to read the comments, but I'm still a bit lost.

tomponline:
> Please would you recap what the issue exactly is? Like what conditions lead to it. A reproducer or a unit test would be great too, although I guess that might be tricky.
> 
> I've tried to read the comments, but I'm still a bit lost.

The original issue has a reproducer https://github.com/canonical/go-dqlite/issues/182

tomponline:
Its not as bad as it was since I added TCP user timeouts in LXD but I still see these sorts of errors for a bit when cluster members go down.

tomponline:
Basically go dqlite keeps offering up a closed connection to be used.

MathieuBordere:
> Please would you recap what the issue exactly is? Like what conditions lead to it. A reproducer or a unit test would be great too, although I guess that might be tricky.
> 
> I've tried to read the comments, but I'm still a bit lost.

I'll try to come up with a reproducer in the unit tests, I'll understand it better myself too then.  The crux of the problem is that as long as a connection is not marked with `ErrBadCon` then the go sql driver will not try a new connection see e.g here in `go 1.18` https://cs.opensource.google/go/go/+/refs/tags/go1.18.4:src/database/sql/sql.go;drc=2580d0e08d5e9f979b943758d3c49877fb2324cb;l=1709

The `SessionResetter` interface allows us to mark a connection with `ErrBadCon` when a connection is taken from the connection pool. In this case e.g. a connection to an offline leader is taken from the pool, `SessionResetter` will try to detect EOF and close the connection and immediately try a new one.

tomponline:
> hi @tomponline , I'm having trouble running the LXD test suite on one of Stéphane's MAAS machines with this branch. It's probably more related to me doing something wrong. Could you run the LXD test suite with the contents of this branch please? Might be quicker.

Testing this now....

tomponline:
@MathieuBordere clean test run here using:

```
cd ~
go work init
go work use ./lxd/
go work use ./go-dqlite/ # With this checked out on your branch
cd lxd
make # I could see changes in ~/go-dqlite being compiled
```

MathieuBordere:
I'm okay to merge this if @cole-miller or @freeekanayaka don't have objections, we can always revert if something unexpected would happen. This branch also ran fine with the jepsen suite.

driver/driver.go
```diff
@@ -787,6 +787,12 @@ type unwrappable interface {
 	Unwrap() error
 }
 
+// TODO driver.ErrBadConn should not be returned when there's a possibility that
+// the query has been executed. In our case there is a window in protocol.Call
```
freeekanayaka, line 791: It probably goes without saying, but just to be sure we are on the same page: this issue should be dealt with by implementing session support in the dqlite state machine, as per paragraph 6.3 of the raft dissertation.

Until then, I'm fine converting `io.EOF` into `driver.ErrBadConn` if it turns out that it solves more problems than it creates (i.e. the chance of returning `driver.ErrBadConn` also if a query was executed is low, while the chance that converting `io.EOF` into `driver.ErrBadConn` solves actual real world issues by retrying is high).

Still, I would have been happier if there was an end-to-end test for this change, showing that we at least precisely understand and can reproduce the exact sequence of events around the issue. With such a test in place it would be easier to avoid regressions later down the road, especially when we'll implement sessions as per 6.3.

freeekanayaka (approved):

MathieuBordere:
I'll merge this tomorrow morning (CET) and make a new release then.

MathieuBordere:
@tomponline https://github.com/canonical/go-dqlite/releases/tag/v1.20.0

tomponline:
> @tomponline https://github.com/canonical/go-dqlite/releases/tag/v1.20.0

Many thanks! :)

PR #194, opened 2022-08-04
driver: Prevent overflowing the amount of parameters

MathieuBordere:
Don't silently overflow number of parameters in tuple.

### DIFF
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -17,6 +17,7 @@ package driver
 import (
 	"context"
 	"database/sql/driver"
+	"fmt"
 	"io"
 	"net"
 	"reflect"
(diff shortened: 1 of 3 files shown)


tomponline (approved):
Thanks!

driver/driver.go
```diff
@@ -371,6 +372,10 @@ func (c *Conn) Prepare(query string) (driver.Stmt, error) {
 
 // ExecContext is an optional interface that may be implemented by a Conn.
 func (c *Conn) ExecContext(ctx context.Context, query string, args []driver.NamedValue) (driver.Result, error) {
+	if len(args) > 255 {
```
tomponline, line 375: Can you change this to a constant somewhere so its easy to change when the underlying limitation is fixed?

tomponline (approved):

PR #196, opened 2022-08-22
internal/bindings/server: don't use bad dqlite_node

cole-miller:
If dqlite_node_create fails, the `struct dqlite_node *` argument is set to NULL, so we shouldn't try to dereference it.

Discovered while investigating https://github.com/canonical/raft/issues/228.

### DIFF
--- a/internal/bindings/server.go
+++ b/internal/bindings/server.go
@@ -85,6 +85,12 @@ import (
 	"github.com/canonical/go-dqlite/internal/protocol"
 )
 
+const (
+	DQLITE_ERROR = 1
+	DQLITE_MISUSE = 2
+	DQLITE_NOMEM = 3
+)
+
 type Node struct {
 	node   *C.dqlite_node
 	ctx    context.Context
@@ -128,8 +134,16 @@ func NewNode(ctx context.Context, id uint64, address string, dir string) (*Node,
 	defer C.free(unsafe.Pointer(cdir))
 
 	if rc := C.dqlite_node_create(cid, caddress, cdir, &server); rc != 0 {
-		errmsg := C.GoString(C.dqlite_node_errmsg(server))
-		return nil, fmt.Errorf("%s", errmsg)
+		var errmsg string
+		switch rc {
+		case DQLITE_ERROR:
+			errmsg = "internal dqlite error"
+		case DQLITE_MISUSE:
+			errmsg = "misuse of dqlite API"
+		case DQLITE_NOMEM:
+			errmsg = "out of memory"
+		}
+		return nil, fmt.Errorf("failed to create node: %s", errmsg)
 	}
 
 	node := &Node{node: (*C.dqlite_node)(unsafe.Pointer(server))}


internal/bindings/server.go
```diff
@@ -128,8 +128,7 @@ func NewNode(ctx context.Context, id uint64, address string, dir string) (*Node,
 	defer C.free(unsafe.Pointer(cdir))
 
 	if rc := C.dqlite_node_create(cid, caddress, cdir, &server); rc != 0 {
-		errmsg := C.GoString(C.dqlite_node_errmsg(server))
-		return nil, fmt.Errorf("%s", errmsg)
+		return nil, fmt.Errorf("failed to create node")
```
tomponline, line 131: We normally start error messages with first letter capitalized in lxd, is this something we can do in go-dqlite too or is there already a precedent for lower case first errors? Thanks

tomponline: Is there another way to get the error message out of dqlite? As otherwise the error itself isn't too useful.

cole-miller: Not sure, there are other uncapitalized Errorf calls in that file but that might be an oversight. I'll see what the majority usage in the project is.

tomponline: thanks!

tomponline: Does rc represent an error code we can include in the error message perhaps?

freeekanayaka: > Does rc represent an error code we can include in the error message perhaps?

Indeed, it does.

cole-miller: Yes, it does -- there are two possible values in current dqlite I believe. I'll amend to include that information.

tomponline: Thanks

freeekanayaka: > Not sure, there are other uncapitalized Errorf calls in that file but that might be an oversight. I'll see what the majority usage in the project is.

I believe the convention in golang is usually to have uncapitalized error messages, and go-dqlite is following that. The LXD convention presumably comes from libc/strerror (wild guess). YMMV

freeekanayaka: The fact that `C.dqlite_node_errmsg(server)` is not returning a proper error message is a bug. It might be a bit laborious but it'd be nice to fix that, since it's going to provide much better diagnostic than the `rc` number alone.

cole-miller: @freeekanayaka I don't think it's a problem with dqlite_node_errmsg, it's just that we're calling it on NULL, since the `struct dqlite_node *` argument is cleared here: https://github.com/canonical/dqlite/blob/54d8a0b67a95249e20a0567a4c5f994914bf1b77/src/server.c#L162

cole-miller:
Pushed an amended commit that includes a bit more detail in the error message.

@tomponline there's substantial numbers of both capitalized and uncapitalized Errorf calls in go-dqlite -- maybe it would be a good idea to normalize on one style in a separate PR?

internal/bindings/server.go
stgraber: Right, since the dqlite_node_create function clears the entire struct on failure, it makes little sense in calling dqlite_node_errmsg on the now NULL struct.

tomponline (commented):
Lgtm!

tomponline:
> @tomponline there's substantial numbers of both capitalized and uncapitalized Errorf calls in go-dqlite -- maybe it would be a good idea to normalize on one style in a separate PR?

@cole-miller yeah sounds good, I'm happy with @freeekanayaka point of using lower case errors in the go-dqlite library (as they will be wrapped in LXD with a upper case first contextual error anyway), but having it consistent here would be good.

PR #197, opened 2022-08-23
style: Normalize error strings

cole-miller:
We use the conventional Go style: no initial capitalization, no final
punctuation. See
https://github.com/golang/go/wiki/CodeReviewComments#error-strings.

As discussed in #196.


tomponline (approved):
Thanks!

PR #200, opened 2022-09-07
app: Don't transfer leadership more than once

cole-miller:
In the `Handover` method of `App`, after successfully transferring leadership to another node, we shouldn't continue looping over other servers and trying to transfer to them. Spotted by @MathieuBordere.


tomponline (approved):
Good catch! :)

tomponline:
@MathieuBordere please could you do a new release of this package so we can import it into LXD?

PR #201, opened 2022-09-07
github: Run go test in CI

cole-miller:
We already run `go test` on Launchpad, but not for pull requests. Not sure whether there's a reason for that, so I'm opening this PR to facilitate discussion.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -44,6 +44,7 @@ jobs:
         deadcode
         export GO_DQLITE_MULTITHREAD=1
         overalls -project ${{ github.workspace }} -covermode=count -- -tags libsqlite3 -timeout 240s
+        go test -v ./...
         VERBOSE=1 ./test/dqlite-demo.sh
         VERBOSE=1 ./test/roles.sh
         VERBOSE=1 ./test/recover.sh


tomponline (commented):
Is there a reason you dont use go test go-dqlite/...?

tomponline:
@markylaing I would appreciate your thoughts on this one given  youve done a lot of our github ci stuff lately. Thanks

cole-miller:
I'm not super familiar with the Go command-line syntax, so I just copied what Launchpad was doing. I'll simplify it, one sec...

tomponline:
> > Is there a reason you dont use go test go-dqlite/...?
> 
> I'm not super familiar with the Go command-line syntax, so I just copied what Launchpad was doing. I'll simplify it, one sec...

Thanks you could take a look at what we do in the lxd github repo, as that way we won't need to remember to add each sub package to the list. Probably worth updating launchpad too otherwise things may be missed there also. Thanks

tomponline (commented):
Why vet off btw?

cole-miller:
> Why vet off btw?

That's also as in Launchpad -- I wanted to keep the two in sync, but I can
remove `-vet=off` in both places if you approve (and if you can point me to
where the definitions for the go-dqlite Launchpad build jobs live -- I
haven't been able to find them).

stgraber:
https://github.com/canonical/dqlite-ppa

tomponline:
Does go test pass with vet off removed? If so I see no reason to leave it disabled. As for launchpad I'm not sure but agree they should be in sync. @stgraber can you point us to where the launchpad build scripts are? Thanks

tomponline:
> https://github.com/canonical/dqlite-ppa

Thanks!

cole-miller:
> Does go test pass with vet off removed?

Yes! Updated this PR. As for Launchpad, as far as I can tell, the arguments passed to `go test` aren't defined by us at all, they're managed by debian's packaging tools, which have their own idea of how to build and test Go projects.

MathieuBordere:
`overalls` runs the tests, see [here](https://github.com/go-playground/overalls).

The following line in the .github actions file takes care of it.
overalls -project ${{ github.workspace }} -covermode=count -- -tags libsqlite3 -timeout 240s

```
$ overalls -help

usage: overalls -project=[path] -covermode[mode] OPTIONS -- TESTOPTIONS

overalls recursively traverses your projects directory structure
running 'go test -covermode=count -coverprofile=profile.coverprofile'
in each directory with go test files, concatenates them into one
coverprofile in your root directory named 'overalls.coverprofile'
```

markylaing:
I had never heard of overalls before! Overall the test/lint actions in this repo look fine. We could use [golangci-lint](https://golangci-lint.run/usage/linters/) to add a few more linters like in LXD (see the [static analysis target in the Makefile](https://github.com/lxc/lxd/blob/a1e23986ad42a1a4d9d856b694fb670f805217a2/Makefile#L245-L269) and the [golangci-lint configuration file](https://github.com/lxc/lxd/blob/a1e23986ad42a1a4d9d856b694fb670f805217a2/.golangci.yml)).

cole-miller:
Thanks @MathieuBordere, looks like we do get test failure reports from overalls, so I'll revert this.

PR #208, opened 2022-09-22
Don't require number of standbys to be odd

cole-miller:
It makes sense to require that the configured number of voters be odd, because that avoids ties in raft elections. But I don't see any reason to require that the number of standby nodes be odd. Please correct me if there's something I'm missing!

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -220,11 +220,6 @@ func New(dir string, options ...Option) (app *App, err error) {
 		return nil, fmt.Errorf("invalid voters %d: must be an odd number greater than 1", o.Voters)
 	}
 
-	if o.StandBys%2 == 0 {
-		stop()
-		return nil, fmt.Errorf("invalid stand-bys %d: must be an odd number", o.StandBys)
-	}
-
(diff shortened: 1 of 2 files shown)


stgraber:
Let's see if @freeekanayaka can think of a reason for that, but personally I don't :)

freeekanayaka:
> Let's see if @freeekanayaka can think of a reason for that, but personally I don't :)

I can't think of any indeed right now. Not sure why it was done like that in the first place, hopefully we're not missing something :)

freeekanayaka (approved):

MathieuBordere:
[This](https://github.com/canonical/go-dqlite/commit/e64d8157b0e0dccbdef127101b57a5a889fb4ba2) is the commit that introduced that line, before it required the number of `Standbys` to be even and I don't really see any logic changes based on going from an even to  an odd requirement. It looks fine I think.

MathieuBordere:
We could require the number of `Standbys` to be greater than 0, but that's maybe artificially restrictive.

PR #214, opened 2022-11-28
Disk mode

MathieuBordere:
Testing

Issue #215, opened 2022-11-28
CI Failure: bind address already in use


MathieuBordere:
Investigate CI Failure:

```
2022/11/28 16:45:03 --- FAIL: TestNew_PristineDefault (0.06s)
2022/11/28 16:45:03     app_test.go:29: 
2022/11/28 16:45:03         	Error Trace:	app_test.go:1045
2022/11/28 16:45:03         	            				app_test.go:944
2022/11/28 16:45:03         	            				app_test.go:29
2022/11/28 16:45:03         	Error:      	Received unexpected error:
2022/11/28 16:45:03         	            	listen to 127.0.0.1:9000: listen tcp 127.0.0.1:9000: bind: address already in use
```

Pops up quite a bit [here](https://github.com/canonical/go-dqlite/actions/runs/3566737312/jobs/5993725687) and [here](https://github.com/canonical/go-dqlite/actions/runs/3566737312/jobs/5993561791) for example/

PR #218, opened 2022-12-08
Disk mode

MathieuBordere:
Reopened after accidental merge of #214 and revert in #217 .


MathieuBordere:
There are a couple of CI runs failing with instances of #215 , I can't imagine these code changes to be the cause of this though.

tomponline (approved):

PR #224, opened 2023-03-06
Add go 1.19

MathieuBordere:
This PR introduces:

 - Update SimpleTLSConfig: The current config doesn't seem to be compatible with `go 1.19.6` making connections fail with `write handshake: local error: tls: bad record MAC` see e.g. [here](https://github.com/canonical/go-dqlite/actions/runs/4342334566/jobs/7583025948), don't know root cause yet and it's probably not a good idea to wait for a new go release if it would indeed be a bug in go (which it is probably not). 
  The config in this PR fixes the issue. Anyone has an opinion on the proposed config in this PR?
- Run CI on Ubuntu 22.04.
- Disable running CI on Ubuntu 18.04
- Add go versions until 1.19 to CI.
- Fix some issues when upgrading to `go 1.19.x` 
  - removing `deadcode`, `overalls`
  - used buffered channel for os.Signal
  - make sure there's only 1 Output comment at end of function

edit: Just removing the curve preferences from the config seems to also fix the issue, this is probably a safer bet than also forcing TLS 1.3

### DIFF
--- a/app/example_test.go
+++ b/app/example_test.go
@@ -28,7 +28,6 @@ func Example() {
 	}
 
 	fmt.Printf("0x%x %s\n", node.ID(), node.Address())
-	// Output: 0x2dc171858c3155be 127.0.0.1:9001
 
 	if err := node.Close(); err != nil {
 		return
@@ -92,7 +91,7 @@ func ExampleWithCluster() {
 	}
 
 	fmt.Println(node1.ID() != node2.ID(), node1.ID() != node3.ID(), node2.ID() != node3.ID())
-	// Output: true true true
+	// true true true
 
 	// Restart the third node, the only argument we need to pass to
 	// app.New() is its dir.
--- a/cmd/dqlite-benchmark/dqlite-benchmark.go
+++ b/cmd/dqlite-benchmark/dqlite-benchmark.go
@@ -44,7 +44,7 @@ const (
 )
 
 func signalChannel() chan os.Signal {
-	ch := make(chan os.Signal)
+	ch := make(chan os.Signal, 32)
 	signal.Notify(ch, unix.SIGPWR)
 	signal.Notify(ch, unix.SIGINT)
 	signal.Notify(ch, unix.SIGQUIT)
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -37,7 +37,6 @@ jobs:
         sudo add-apt-repository ppa:dqlite/dev -y
         sudo apt update
         sudo apt install -y golint libsqlite3-dev libuv1-dev liblz4-dev libraft-dev libdqlite-dev
-        go get github.com/tsenart/deadcode
         go get github.com/go-playground/overalls
 
     - name: Build & Test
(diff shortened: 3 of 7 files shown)


MathieuBordere:
@tomponline @cole-miller Mostly just asking feedback on the updated `SimpleTLSConfig`, will do more research myself on possible consequences, but your input is very welcome.

MathieuBordere:
My worry is a bit that current clients could use certificates that are incompatible with TLS 1.3

I could do a sanity check on the provided cert and certpool, like setup a server and client with the provided config and see if I can setup a connection. If not, revert to TLS 1.2 or err out.

tomponline:
CC @simondeziel may have a comment on this .

cmd/dqlite-benchmark/dqlite-benchmark.go
```diff
@@ -44,7 +44,7 @@ const (
 )
 
 func signalChannel() chan os.Signal {
-	ch := make(chan os.Signal)
+	ch := make(chan os.Signal, 32)
```
tomponline, line 47: Why 32 out of interest?

tomponline:
> My worry is a bit that current users could use certificates that are incompatible with TLS 1.3
> 
> I could do a sanity check on the provided cert and certpool, like setup a server and client with the provided config and see if I can setup a connection. If not, revert to TLS 1.2 or err out.

I think I would be leaning towards removing as many of the explicitly set options as possible and relying on the defaults from Go as they change from version to version (thinking MinVersion, MaxVersion, CurvePreferences, PreferServerCipherSuites (deprecated), CipherSuites).

And then make that clear in the comment, and if a user needs specific config then they can create a custom config.

cmd/dqlite-benchmark/dqlite-benchmark.go
tomponline: Its fine, just I always wonder about "magic numbers" :)

simondeziel (commented):
I quite like the simplification for the TLS server and client.

LXD 5.0+ requires [TLS 1.3 unless LXD_INSECURE_TLS is set](https://github.com/lxc/lxd/blob/master/shared/network.go#L71) and that applies to both the server and the client.

MathieuBordere:
@freeekanayaka maybe you have an opinion on the TLS configs?

app/example_test.go
```diff
@@ -28,7 +28,6 @@ func Example() {
 	}
 
 	fmt.Printf("0x%x %s\n", node.ID(), node.Address())
-	// Output: 0x2dc171858c3155be 127.0.0.1:9001
```
freeekanayaka, line 31: Why was this change needed? Has format of testable examples changed in Go 1.19? I would be very surprised.

freeekanayaka:
Looks good to me. Those settings were the "recommended" ones in terms of tightening TLS parameters, but it seems reasonable to just use Go's defaults, which might be a bit more permissive, but clearly good enough.

app/example_test.go
MathieuBordere: go 1.19 only allows a single Output comment block at the end of a function, and there is already one at the end, see e.g. [this](https://github.com/canonical/go-dqlite/actions/runs/4341772301/jobs/7581855576) test run.

```
# github.com/canonical/go-dqlite/app_test
Error: app/example_test.go:31:2: there can only be one output comment block per example
Error: app/example_test.go:95:2: output comment block must be the last comment block
```

PR #225, opened 2023-03-07
fix: ensure creating a New app is thread safe

djgilcrease:
Using a global int that is not atomic causes issues if you try to create new drivers in go routines as happens in testing.
Fixes two other data races found in testing


app/app.go
```diff
@@ -707,4 +708,4 @@ func (a *App) error(format string, args ...interface{}) {
 	a.log(client.LogError, format, args...)
 }
 
-var driverIndex = 0
+var driverIndex = atomic.Int64{}
```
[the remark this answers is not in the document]
djgilcrease: Fixed, and signed the CLA

djgilcrease:
Added a test to show the data race and verified the test passes after the change

djgilcrease:
@cole-miller can I drop the go1.13 tests since the go.mod specifies go1.14? That will allow me to use the *testing.T.Cleanup function for test cleanup which is more consistent than a deferred function.

djgilcrease:
> I wouldn't object to requiring Go 1.14 or later, although I don't regard that line in go.mod as indicating that we don't support Go 1.13. @tomponline, @MathieuBordere, your thoughts?

for now, I just switched back to the `defer func(){...}` method, but according to https://go.dev/doc/modules/gomod-ref#go-syntax that is supposed to indicate the minimum version of go supported.

djgilcrease:
Ok, I dropped go1.13 and added go1.20 since it is out now. Also found a way to fix the data race in `connectAttemptAll` without needing the mutex lock by just copying the slice before sorting it.

internal/protocol/connector.go
```diff
@@ -119,13 +119,18 @@ func (c *Connector) Connect(ctx context.Context) (*Protocol, error) {
 // Make a single attempt to establish a connection to the leader server trying
 // all addresses available in the store.
 func (c *Connector) connectAttemptAll(ctx context.Context, log logging.Func) (*Protocol, error) {
+	// Wait here until we can lock the servers data
+	for {
+		if c.lock.TryLock() {
```
tomponline: I am concerned about this, it looks like a busy wait to me? What is this for? What was wrong with using c.lock.Lock()?

tomponline (changes requested):
@djgilcrease I started reviewing this, but then realised half way through that this PR hasn't been rebased to the "final" version yet (apologies for the out of date comments). 

Please can you rebase the PR so that you only have logical commits representing the "final" version you want reviewed, rather than showing the evolution of the PR. Then I'll take a look again, thanks.

tomponline:
> I wouldn't object to requiring Go 1.14 or later, although I don't regard that line in go.mod as indicating that we don't support Go 1.13. @tomponline, @MathieuBordere, your thoughts?

For what its worth, LXD is now requiring Go 1.18 as minimum version, so unless you have specific requirements for less than that, I don't see an issue with matching LXD's minimum Go version.

djgilcrease:
Rebased to be a single logical commit, and updated the min go version to 1.18 to match LXD

internal/protocol/connector.go
```diff
@@ -116,11 +116,13 @@ func (c *Connector) Connect(ctx context.Context) (*Protocol, error) {
...
 func (c *Connector) connectAttemptAll(ctx context.Context, log logging.Func) (*Protocol, error) {
-	servers, err := c.store.Get(ctx)
+	storedServers, err := c.store.Get(ctx)
 	if err != nil {
 		return nil, errors.Wrap(err, "get servers")
 	}
-
+	// Create a copy of the stored servers to get around the data race
+	servers := make([]NodeInfo, len(storedServers))
+	copy(servers, storedServers)
```
tomponline, line 125: How do you stop the race occurring here?

tomponline: I would have thought this would need to be covered by a mutex for all operations of c.store.Get()

MathieuBordere:
Changing the minimum go version will break our ppa package on focal, where the go version provided by apt is still 1.13 if I'm not mistaken, so can't really agree with that change.

tomponline:
> Changing the minimum go version to 1.18 will break our ppa package on focal, where the go version provided by apt is still 1.13 if I'm not mistaken, so can't really agree with that change.

Makes sense, in that case the min version must remain 1.13.

tomponline:
> > > Changing the minimum go version to 1.18 will break our ppa package on focal, where the go version provided by apt is still 1.13 if I'm not mistaken, so can't really agree with that change.
> > 
> > 
> > Makes sense, in that case the min version must remain 1.13.
> 
> Min version was already go1.14 in go.mod

Sounds like it needs to be dropped back down to 1.13 though.

internal/protocol/connector.go
djgilcrease: The data race is actually in the slice sort and it appears to be caused by `Set` being called on the store in the middle of the sort so operating on a copy of the slice fixes that. The lock does not fix the issue because the call to `Set` is done outside of the connector which will allow the slice to still be modified in the middle of the sort.

tomponline: Thanks for the clarification. Does the race still exist though if the `storedServers` is modified during `copy`?

djgilcrease: I ran the test that was triggering it with `-count 1000` and could not reproduce it. Without the `copy` it reproduced on 90% of runs with `-count 1` and 100% of the time if I had `-count 2`

tomponline: OK, its certainly an improvement, and won't prevent this PR from being merged, but I don't think its fully solved it.

tomponline (approved):
I'm happy with this if @MathieuBordere and @cole-miller are :)

internal/protocol/connector.go
djgilcrease: Ya, likely a better way to solve it is to have a lock in the store and make a copy in the get method and return that copy

djgilcrease:
Updated this to move the lock/copy to the store for the `InmemNodeStore` store and add the copy to the `YamlNodeStore` since it already had a lock.

PR #227, opened 2023-03-07
driver: Add some additional tracing

### LINKED ISSUE #226: Log Start and End of requests
Ensure to [log](https://github.com/canonical/go-dqlite/blob/7d04094a32ae2afb880086c9c6acf985c779c1b1/driver/driver.go#L422) when receiving a request and when responding to it.

CC: @SimonRichardson

cole-miller:
Closes #226. @SimonRichardson, does this suit your needs?

### DIFF
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -351,11 +351,18 @@ func (c *Conn) PrepareContext(ctx context.Context, query string) (driver.Stmt, e
 
 	protocol.EncodePrepare(&c.request, uint64(c.id), query)
 
-	if err := c.protocol.Call(ctx, &c.request, &c.response); err != nil {
+	var start time.Time
+	if c.tracing != client.LogNone {
+		start = time.Now()
+	}
+	err := c.protocol.Call(ctx, &c.request, &c.response);
+	if c.tracing != client.LogNone {
+		c.log(c.tracing, "%.3fs request prepared: %q", time.Since(start).Seconds(), query)
+	}
+	if err != nil {
 		return nil, driverError(c.log, err)
 	}
 
-	var err error
 	stmt.db, stmt.id, stmt.params, err = protocol.DecodeStmt(&c.response)
 	if err != nil {
 		return nil, driverError(c.log, err)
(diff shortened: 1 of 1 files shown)


driver/driver.go
```diff
@@ -351,11 +351,17 @@ func (c *Conn) PrepareContext(ctx context.Context, query string) (driver.Stmt, e
...
 	protocol.EncodePrepare(&c.request, uint64(c.id), query)
 
-	if err := c.protocol.Call(ctx, &c.request, &c.response); err != nil {
+	if c.tracing != client.LogNone {
+		c.log(c.tracing, "request-start prepare %s", query)
+	}
+	err := c.protocol.Call(ctx, &c.request, &c.response);
+	if c.tracing != client.LogNone {
+		c.log(c.tracing, "request-end prepare %s", query)
+	}
```
SimonRichardson, line 360: I'll be honest, I'm not too fond of this design pattern. It makes it very difficult to parse the logs, not only for manual inspection but also for automated inspection. Those two log lines aren't guaranteed to follow each other, so it will require looking for the start and the end. Additionally, if you have two queries that are the same, but are interleaved, how do you know the correct ordering?

What I was hoping for was something more akin to the following. Essentially, have the ability to calculate the duration of a request if tracing is enabled.

```suggestion
	var start time.Time
	if c.tracing != client.LogNone {
		start = time.Now()
	}
	err := c.protocol.Call(ctx, &c.request, &c.response);
	if c.tracing != client.LogNone {
		c.log(c.tracing, "request prepare %s: %d", query, start.Since(time.Now))
	}
```

@manadart or @jameinel thoughts?

jameinel: Long term, we'll need something much better than just log messages scattered about to figure out where we are spending time, what queries are slow, etc. Either we need to do this in the Juju layer, or we get changes at the dqlite layer to provide structured data. (what queries, what tables, how long as a Duration not as a string, etc)
I do agree that Simon's proposal is better than separate log messages. Especially since the log messages don't appear to have explicit context to match them together. (You could, for example, start generating a unique identifier, and shipping that in the log message, so that you could match up the pairs. We do this in juju logs with a connection id and a request id, so every connection has a known set of requests and responses to those requests)

Personally I also like to put the timestamps at the beginning, especially if the query is freeform text. so:
```
c.log(c.tracing, "%s request prepared: %q", time.Since(start), query)
```
(note that you also don't want start.Since() since that would be a negative Duration :)

And many do like that "%s" for formatting a time.Duration will add us/ms/s/etc on its own. I would say that I actually kind of prefer:
```
c.log(c.tracing, "%.3fs request prepared: %q", time.Since(start).Seconds(), query)
```

That gives millisecond resolution (and if it is faster than that, I probably don't care about it), and means that all the times you get are in the same units. (so you don't see 500us and worry that it is a problem vs the 5ms that you are getting elsewhere)

SimonRichardson (approved):

Issue #228, opened 2023-03-09
Handover tests are flaky


MathieuBordere:
e.g.

```
2023-03-08T17:57:29.5023570Z     app_test.go:515: 
2023-03-08T17:57:29.5024022Z         	Error Trace:	app_test.go:515
2023-03-08T17:57:29.5024518Z         	Error:      	Received unexpected error:
2023-03-08T17:57:29.5025423Z         	            	demote ourselves: a configuration change is already in progress (5)
2023-03-08T17:57:29.5025997Z         	Test:       	TestHandover_TransferLeadership
2023-03-08T17:57:29.5026503Z --- FAIL: TestHandover_TransferLeadership (3.22s)
```

I think this is happening since a change in raft, where a new leader will only commit log entries from its current term (and not previous terms) by counting replicas.  The transferred to leader still needs to commit the last configuration change because it becomes leader before receiving the latest commit index from the previous leader. The latest log entry is a configuration change, so [this](https://github.com/canonical/go-dqlite/blob/a51f109f7ae88d7978aa172eda07144abd979745/app/app.go#L376) step fails in the tests. The old leader is not able to demote (become non-voter) itself. 

This is problematic in a real cluster though (e.g. with 2 nodes), if the old leader goes offline before demoting itself, we're left with a 2 node cluster with 2 voters of which only 1 is online and the cluster is unavailable.

PR #229, opened 2023-03-09
app: Retry Handover until ctx times out.

MathieuBordere:
Fixes https://github.com/canonical/go-dqlite/issues/228 .

In the issue above the demotion of the old leader fails because the new transferred to leader is still applying it's no-op barrier entry at the start of its term, resulting in an uncommitted configuration change while that barrier has not been committed. The retry logic works around this, because the barrier will eventually commit and the demotion can continue.

This reminds us of the issue that @cole-miller brought up a while ago, that when trying to update the raft configuration we would also need a barrier.

I'm okay with the semantics that a config change can fail because another one is ongoing and that a client should just retry it.


freeekanayaka:
Hi, I'm a bit busy these days. Please go ahead, I'll see if I can catch up later on.

PR #230, opened 2023-03-10
refactor: replace `github.com/ghodss/yaml` with `sigs.k8s.io/yaml`

Juneezee:
At the time of making this commit, the package `github.com/ghodss/yaml` is no longer actively maintained.

`sigs.k8s.io/yaml` is a permanent fork of `ghodss/yaml` and is actively maintained by Kubernetes SIG.


tomponline:
> Thanks for the PR and for bringing up the maintenance status of `github.com/ghodss/yaml`. Since our YAML needs are so simple, perhaps we could get away with just using `gopkg.in/yaml.v2` (or `v3`) directly and eliminating a dependency?

That would be preferable.

Juneezee:
> @Juneezee, would you be up for reworking this PR to use `gopkg.in/yaml.v3`?

Done!

Juneezee:
> I just ran some tests locally and it seems that -- unlike `github.com/ghodss/yaml` -- `gopkg.in/yaml.v3` does some case conversion by default

Yup, I noticed the same too. The tests are failing with both `gopkg.in/yaml.v2` and `gopkg.in/yaml.v3` (https://github.com/Juneezee/go-dqlite/actions/runs/4390622459). 

`gopkg.in/yaml.v2` and `gopkg.in/yaml.v3` marshal and unmarshal using the field name lowercased as default key if we don't explicitly specify a `yaml` field tag. (https://pkg.go.dev/gopkg.in/yaml.v3#Marshal)

MathieuBordere:
seems to hit a segfault on s390x https://launchpadlibrarian.net/655817390/buildlog_ubuntu-focal-s390x.go-dqlite_1.11.7+git5-g394129c~focal1_BUILDING.txt.gz

PR #243, opened 2023-04-11
internal/protocol: Expose 32-bit parameter count support

### LINKED ISSUE #210: Support for new-style requests with more parameters
I've implemented server-side support for versions of the EXEC, EXEC_SQL, QUERY, and QUERY_SQL requests that can include more statement parameters in canonical/dqlite#407. We'll want to implement client-side support for these in go-dqlite. The higher-level interface in the client package will need to take into account that older dqlite servers won't handle the new-style requests properly (they will try and presumably fail to parse them in the old format), perhaps with some kind of graceful fallback to the more limited old-style requests.

cole-miller:
Closes #210

### DIFF
--- a/driver/driver_test.go
+++ b/driver/driver_test.go
@@ -226,7 +226,7 @@ func TestStmt_Exec(t *testing.T) {
 	assert.NoError(t, conn.Close())
 }
 
-func TestStmt_ExecTooManyParams(t *testing.T) {
+func TestStmt_ExecManyParams(t *testing.T) {
 	drv, cleanup := newDriver(t)
 	defer cleanup()
 
@@ -244,15 +244,15 @@ func TestStmt_ExecTooManyParams(t *testing.T) {
 
 	require.NoError(t, stmt.Close())
 
-	stmt, err = conn.Prepare("INSERT INTO test(n) VALUES " + strings.Repeat("(?), ", 255) + " (?)")
+	stmt, err = conn.Prepare("INSERT INTO test(n) VALUES " + strings.Repeat("(?), ", 299) + " (?)")
 	require.NoError(t, err)
 
-	values := make([]driver.Value, 256)
+	values := make([]driver.Value, 300)
 	for i := range values {
 		values[i] = int64(1)
 	}
 	_, err = stmt.Exec(values)
-	require.Errorf(t, err, "too many parameters (256) max = 255")
+	require.NoError(t, err)
 
 	require.NoError(t, stmt.Close())
 	assert.NoError(t, conn.Close())
@@ -304,7 +304,7 @@ func TestStmt_Query(t *testing.T) {
 	assert.NoError(t, conn.Close())
 }
 
-func TestStmt_QueryTooManyParams(t *testing.T) {
+func TestStmt_QueryManyParams(t *testing.T) {
 	drv, cleanup := newDriver(t)
 	defer cleanup()
 
@@ -322,15 +322,15 @@ func TestStmt_QueryTooManyParams(t *testing.T) {
 
 	require.NoError(t, stmt.Close())
 
-	stmt, err = conn.Prepare("SELECT n FROM test WHERE n IN (" + strings.Repeat("?, ", 255) + " ?)")
+	stmt, err = conn.Prepare("SELECT n FROM test WHERE n IN (" + strings.Repeat("?, ", 299) + " ?)")
 	require.NoError(t, err)
 
-	values := make([]driver.Value, 256)
+	values := make([]driver.Value, 300)
 	for i := range values {
 		values[i] = int64(1)
 	}
 	_, err = stmt.Query(values)
-	require.Errorf(t, err, "too many parameters (256) max = 255")
+	require.NoError(t, err)
 
 	require.NoError(t, stmt.Close())
 	assert.NoError(t, conn.Close())
@@ -383,7 +383,7 @@ INSERT INTO test (n,t) VALUES (3,'b');
 	assert.NoError(t, conn.Close())
 }
 
-func TestConn_QueryTooManyParams(t *testing.T) {
+func TestConn_QueryManyParams(t *testing.T) {
 	drv, cleanup := newDriver(t)
 	defer cleanup()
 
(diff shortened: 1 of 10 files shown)


freeekanayaka:
Looks good to me. I'm agnostic about support for go-dqlite + old dqlite, although this change might hit some users.

One approach would be to use schema 0 if there are at most 256 parameters, and schema 1 if there are more. In that way the only difference with an old dqlite server would be that the old `go-dqlite` would throw a panic, and this new `go-dqlite` would get an error from the server. It would be a bit of work to implement though since the code doesn't quite support switching between schema versions.

MathieuBordere:
> One approach would be to use schema 0 if there are at most 256 parameters, and schema 1 if there are more. In that way the only difference with an old dqlite server would be that the old `go-dqlite` would throw a panic, and this new `go-dqlite` would get an error from the server. It would be a bit of work to implement though since the code doesn't quite support switching between schema versions.

I would be in favor of that, using schema version 0 when parameter count is <= 255.

internal/protocol/request.go
```diff
@@ -45,27 +45,47 @@ func EncodePrepare(request *Message, db uint64, sql string) {
 	request.putUint64(db)
 	request.putString(sql)
 
-	request.putHeader(RequestPrepare)
+	request.putHeader(RequestPrepare, 0)
 }
 
-// EncodeExec encodes a Exec request.
-func EncodeExec(request *Message, db uint32, stmt uint32, values NamedValues) {
+// EncodeExecV0 encodes a Exec request.
```
TheSignPainter98, line 51: This comment doesn't highlight the difference between `EncodeExecV0` and `EncodeExecV1`---the choice between each is unclear. As this comes up in quite a few places, is it worth putting a doc at the top of the file to establish the convention? Perhaps something like:
```
// Many functions in this file have two versions, distinguished
// by their suffix. Those named FooV0 handle queries which accept
// at most 255 arguments. Those named FooV1 handle queries with
// arbitrarily-many arguments but ...

File: internal/protocol/message.go [hunk 2]
```diff
@@ -229,12 +217,46 @@ func (m *Message) putNamedValues(values NamedValues) {
...
+// Encode the given driver values as binding parameters.
+func (m *Message) putNamedValues(values NamedValues) {
+	l := len(values)
+	if l > 255 {
+		// safeguard, should have been checked beforehand.
+		panic("too many parameters")
+	}
+
+	n := uint8(l) // N of params
+	if n == 0 {
```
TheSignPainter98, line 231: Somehow this early return on `len(values) == 0` would feel more natural sooner. Would it be better to have it just before the `l > 255` check?

The same would hold for `Message.putNamedValues32`

internal/protocol/message.go
```diff
@@ -229,12 +217,46 @@ func (m *Message) putNamedValues(values NamedValues) {
...
+
+	m.putUint8(n)
+	m.putNamedValuesInner(values)
+}
+
+// Encode the given driver values as binding parameters, with a 32-bit
+// parameter count (new format).
+func (m *Message) putNamedValues32(values NamedValues) {
+	l := len(values)
+	if l > (1<<32) - 1 {
```
TheSignPainter98, line 243: Would it be clearer to use `math.MaxUint32` here? Admittedly this would feel strange

internal/protocol/message.go
```diff
@@ -229,12 +217,46 @@ func (m *Message) putNamedValues(values NamedValues) {
...
 		}
 	}
+}
+
+// Encode the given driver values as binding parameters.
+func (m *Message) putNamedValues(values NamedValues) {
+	l := len(values)
+	if l > 255 {
+		// safeguard, should have been checked beforehand.
+		panic("too many parameters")
```
TheSignPainter98, line 227: Admittedly this isn't a convention elsewhere (so likely not an issue for this PR) but would it be better to have `"internal error: ..."` to highlight that the user isn't at-fault for these panics?

driver/driver_test.go
```diff
@@ -252,7 +252,7 @@ func TestStmt_ExecTooManyParams(t *testing.T) {
 		values[i] = int64(1)
 	}
 	_, err = stmt.Exec(values)
-	require.Errorf(t, err, "too many parameters (256) max = 255")
+	require.NoError(t, err)
```
freeekanayaka, line 255: It would be nice to not only check that no error occurred, but that the parameters where actually correctly handled, i.e. 256 rows were inserted, with values ranging from 1 to 255.

driver/driver_test.go
```diff
@@ -330,7 +330,7 @@ func TestStmt_QueryTooManyParams(t *testing.T) {
 		values[i] = int64(1)
 	}
 	_, err = stmt.Query(values)
-	require.Errorf(t, err, "too many parameters (256) max = 255")
+	require.NoError(t, err)
```
freeekanayaka, line 333: Same here. It would be good to do something like:

- insert 300 rows, with values ranging from 1 to 300
- run the SELECT statement in the test
- check that exactly 256 rows are returned, with the expected values

Perhaps it'd be worth using a slightly larger value than 256, say 260 or 270, just to test not only the boundary value but also beyond.

internal/protocol/message.go
freeekanayaka: I don't think it can happen earlier, because the point of the `l > 255` check is to be sure that the `n := uint8(l)` cast is safe.

internal/protocol/response.go
```diff
@@ -15,13 +15,13 @@ func DecodeFailure(response *Message) (code uint64, message string, err error) {
 		e := ErrRequest{}
 		e.Code = response.getUint64()
 		e.Description = response.getString()
-                err = e
```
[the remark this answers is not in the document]
freeekanayaka: Good catch.

internal/protocol/schema.sh
```diff
@@ -1,7 +1,7 @@
 #!/bin/bash
 
 request_init() {
-    cat > request.go <<EOF
+	cat > request.go <<EOF
```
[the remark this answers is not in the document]
freeekanayaka: As @TheSignPainter98 pointed, some tabs are needed because of gofmt.

internal/protocol/message.go
TheSignPainter98: I'm not sure I follow, the `> 255` check makes the `!= 0` check look independent to me; is this not possible?
```go
l := len(values)
if l == 0 {
    return
}
if l > 255 {
    panic("...")
}

m.putUint8(uint8(l))
```

PR #248, opened 2023-05-01
Linker flags

MathieuBordere:
Return an error on Node creation when the runtime version of dqlite is smaller than the version used at compile time.

Consequences:

- We can get rid of the `-Wl,-z,-now` flags to detect API calls that are unsupported by the dqlite version found at runtime.
- We are no longer compatible when a `go-dqlite` finds an older dqlite version at runtime that is 100% compatible but has a lower `MAJOR` or `MINOR` version number.

An alternative would be to bake in the minimum required dqlite version in the the `go-dqlite` source instead of using the `DQLITE_VERSION_NUMBER` macro. That way it's documented and explicit.


freeekanayaka (approved):
I think this is good enough, but of course putting in go-dqlite an explicit minimum dqlite version number which we know we require (and bumping it accordingly when using newly introduced features) would be better, although one more bit to maintain manually.

I'm fine either way, since it seems that in practice this might be mostly relevant for people developing go-dqlite/dqlite itself. For most (production) users the dqlite version that go-dqlite links at runtime will typically be the same used at build time, or higher.

MathieuBordere:
> I think this is good enough, but of course putting in go-dqlite an explicit minimum dqlite version number which we know we require (and bumping it accordingly when using newly introduced features) would be better, although one more bit to maintain manually.
> 
> I'm fine either way, since it seems that in practice this might be mostly relevant for people developing go-dqlite/dqlite itself. For most (production) users the dqlite version that go-dqlite links at runtime will typically be the same used at build time, or higher.

I'll change it to an explicit dqlite version, it's a better solution imo.

Issue #238, opened 2023-03-31
Allow WithExternalConn use any net.Conn instance


Zxilly:
I wonder if that is possible to allow `WithExternalConn` use any `net.Conn`.  

As for now, it assume the connection is a `net.TCPConn` or `tls.Conn`.  And there are many behaviors bind to tcp.

MathieuBordere:
dqlite and the underlying raft library are designed assuming that the transport is reliable, you could use any other reliable `net.Conn`.

Zxilly:
yes,  I'd like to use quic as the transport layer for the dqlite.  But as for now, go-dqlite has a function called extractTcpConn, and it cannot support other kind net.Conn.
It will also try to use syscall to set tcp keepalive params.

MathieuBordere:
Ok, thanks for pointing that out, I'll make an issue out of it.

Zxilly:
And 
https://github.com/canonical/go-dqlite/blob/251e831db80e50bb6a54431e31529d12aa1291c5/app/app.go#L275-L280
will lead to handshake failure on transport other than http.

PR #252, opened 2023-05-04
refactor: allow any net.Conn works with dqlite

Zxilly:
Close: #238 

For now, although dqlite accepts any connection that matches `net.Conn`, it only works if the target is tcp.Conn or tls.Conn.

Also, dqlite assumes that users always use `websocket` to communicate, which again limits the flexibility when used as a library.

This patch maintains the previous behavior for tcp and tls, while allowing other constructs that implement the `net.Conn` interface to work properly.


freeekanayaka:
Not a hard blocker, but it would be nice to add a unit test for this, ideally using quic. I believe it shouldn't be too much difficult.

MathieuBordere:
Thanks for this @Zxilly! Can I ask you to sign the [Canonical CLA](https://ubuntu.com/legal/contributors/agreement) please? It's a one time thing, afterwards you can contribute to any Canonical project.

Zxilly:
I have signed CLA, currently I'm working a unit test.

freeekanayaka:
> Seems quic go can not run with go 1.13. Maybe we should use named pipe？

Ah too bad. Is there an easy way to run that test only with go >= 1.19? If not, then yeah, perhaps using something like named pipe would be ok.

Zxilly:
I split test code to make it only compiled on higher version.

For indirectly dependencies I have no good idea. A possible solution can be require a empty package and replace it with real quic-go. With this, the quic-go won't be part of dqlite dependencies.

freeekanayaka:
Thanks @Zxilly, tests are still failing, but the overall testing approach looks good to me.

Zxilly:
I dive into deps dependencies, seems `quic-go` relies on `golang.org/x/sys@v0.5.0`, which need go 1.17 to compile.
I just commented quic test and add another pipe implementation. If one day dqlite set its go version to 1.17+, it can be re-enabled.

PR #253, opened 2023-05-26
Allow BEGIN CONCURRENT for transactions

SimonRichardson:
The following implements a TxMode for testing purposes. The idea being we would like to see if there is any performance difference in using BEGIN CONCURRENT or not in juju.

The notion is simple, we pass a option through the App and to the driver and based on the value of the option we switch the tx start mode.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -218,6 +218,7 @@ func New(dir string, options ...Option) (app *App, err error) {
 		driver.WithDialFunc(driverDial),
 		driver.WithLogFunc(o.Log),
 		driver.WithTracing(o.Tracing),
+		driver.WithTxMode(o.TxMode),
 	)
 	if err != nil {
 		stop()
--- a/app/options.go
+++ b/app/options.go
@@ -189,6 +189,14 @@ func WithDiskMode(disk bool) Option {
 	}
 }
 
+// WithTxMode sets the custom transaction mode between BEGIN and
+// BEGIN CONCURRENT.
+func WithTxMode(mode client.TxMode) Option {
+	return func(options *options) {
+		options.TxMode = mode
+	}
+}
+
 type tlsSetup struct {
 	Listen *tls.Config
 	Dial   *tls.Config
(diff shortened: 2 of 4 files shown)


freeekanayaka:
As far as I knew `BEGIN CONCURRENT` is not part of the current official SQLite release (iow it's an un-merged experimental branch), or did this change recently?

I did not look at the implementation details of `BEGIN CONCURRENT`, but I'm not entirely sure if it's fully compatible with dqlite's custom VFS, as it is now. Did you try it already?

Also, does juju have a lot of write contention, making you think that `BEGIN CONCURRENT` would make a difference? Probably one way to measure that would be to know how many transactions fail with `SQLITE_BUSY` aka `driver.ErrBusy` (i.e. another transaction is holding the write lock) and have to be retried.

freeekanayaka:
> > or did this change recently?
> 
> No, it's still being maintained in an experimental branch (from the github.com mirror https://github.com/sqlite/sqlite/tree/begin-concurrent).
> 
> > Did you try it already?
> 
> No, this PR was to see if it was and also to gain your thoughts.

I personally think `BEGIN CONCURRENT` is a good idea, I'm not sure why it hasn't been merged (I think it's been around for years). Perhaps a combination of not enough user interest and some risks in the implementation.
 
> > Also, does juju have a lot of write contention
> 
> When we scale up we see a LOT of write contention (more information to follow).
> 
> I'll close this PR for now, as it requires a sqlite branch for this to work. I've got it compiling from the branch, but I don't think we'd want to support this long-term.

I agree it's not great long-term if upstream doesn't merge it. The are a few starts that need to align for `BEGIN CONCURRENT` to be effective and produce a performance boost, the model/transactions need to be crafted in a way that eliminates the overlap between pages being read or written by different transactions. Depending on the specific workload of the app, that might be easy or tricky. Not sure about juju, but perhaps it's worth having a rough evaluation of that aspect even before considering `BEGIN CONCURRENT` at all.

I wouldn't rule out `BEGIN CONCURRENT` completely if it turns out that juju's data model is in principle a good fit for non-overlapping transactions and if scaling turns out to be hard without that.

On the other hand, many applications are not very optimized for efficient SQL transactions (since that becomes a problem only at scale), so perhaps there significant room for improvement even without `BEGIN CONCURRENT`. All that can be done in that way, should be done. I remember implementing various SQL improvements in LXD that made it possible to scale (or at least to scale more).

PR #261, opened 2023-07-17
Export raw SQLite result codes

### LINKED ISSUE #259: Client Error Handling needs improvement
Client errors get returned as `protocol.Error` which is located in the `internal` package, this means that users of the library are unable to use the error type to assert on information contained within it, such as `Code`.

Also, is it possible to map and return the `SQLite3` Error codes as seen in `github.com/mattn/go-sqlite3`'s `sqlite3.Error`?
this would make error handling much more user friendly.

Lastly, using the `RETURNING` statement (autogenerated with new versions of sqlboiler) renders any error returned from dqlite unusable and breaks our applications error handling.
Our choices are pin to an old version of sqlboiler or look into dropping dqlite for another project, neither of these are ideal solutions so it would be great if we could have some kind of user-friendly error handling with exposed constants so that we can explicitly check errors returned.

cole-miller:
Closes #259. My preference here is to export all these constants under their "C names", e.g. SQLITE_CONSTRAINT, to make their raw nature obvious, but I can switch this PR to Go-style names if @freeekanayaka or @MathieuBordere feel strongly about it.

### DIFF
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -45,20 +45,125 @@ type Driver struct {
 // Error is returned in case of database errors.
 type Error = protocol.Error
 
-// Error codes. Values here mostly overlap with native SQLite codes.
+// SQLite result codes.
 const (
-	ErrBusy                = 5
-	ErrBusyRecovery        = 5 | (1 << 8)
-	ErrBusySnapshot        = 5 | (2 << 8)
-	errIoErr               = 10
-	errIoErrNotLeader      = errIoErr | 40<<8
-	errIoErrLeadershipLost = errIoErr | (41 << 8)
-	errNotFound            = 12
+	// Primary result codes.
+	ErrError      =  1
+	ErrInternal   =  2
+	ErrPerm       =  3
+	ErrAbort      =  4
+	ErrBusy       =  5
+	ErrLocked     =  6
+	ErrNomem      =  7
+	ErrReadonly   =  8
+	ErrInterrupt  =  9
+	ErrIoErr      = 10
+	ErrCorrupt    = 11
+	ErrNotFound   = 12
+	ErrFull       = 13
+	ErrCantOpen   = 14
+	ErrProtocol   = 15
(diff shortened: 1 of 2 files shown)


freeekanayaka:
I'd probably use the same naming as go-sqlite3: https://github.com/mattn/go-sqlite3/blob/master/error.go, which is both more Go-ish and something more familiar for folks coming from go-sqlite3 itself.

tomponline:
That would  be my preference too.

freeekanayaka:
> I don't really object to changing the names, but note that go-sqlite3 uses the original C-style names for the integer constants

Do you mean these? https://github.com/mattn/go-sqlite3/blob/v1.14.17/sqlite3.go#L302

If so, it seems that go-sqlite3 uses C-style names only for some kinds of constants, but not for error constants specifically?

> (i.e. what we're exporting here) and Go-style names for vars that are wrapped in its special ErrNo type, which I guess exists just so that something can be returned that implements the Error interface. For go-dqlite we already have a custom Error type that fills that role, packing up both the integer result code and the string description; arguably by using the C-style names we'd be sticking closer to what go-sqlite3 does.

FWIW we also already have Go-style names for some of the error constants: https://github.com/canonical/go-dqlite/blob/beebd0121cfa366ebf3cbb9cf9e807af812aa38e/driver/driver.go#L49 so one natural thing to do in my view would be to add the missing ones.

The way I see it, using Go-style names would be 1) consistent with what we already do 2) consistent with what go-sqlite3 does for error constants 3) consistent with Go naming convention in general.

I'd find it odd to have both `driver.ErrBusy` and `driver.SQLITE_BUSY`. In my mind it's actually kind of good to not have the `SQLITE` prefix, because the idea is that this error set is a super-set of SQLite errors, whose values and semantics is mostly the same but that can have subtle differences (which should be ideally documented, although they might not be at the moment).

MathieuBordere:
@cole-miller I'm okay to merge this, does this need anything more?

PR #263, opened 2023-07-18
Don't bail out early in roles adjustment

cole-miller:
Consider a cluster of three dqlite nodes, all of which are voters. If one of these nodes goes offline, we'd like for the leader to eventually demote it to spare during the periodic roles adjustment. This PR fixes a quirk of the `Adjust` function that prevented that demotion from taking place. Specifically, we hit the early return here after noticing that there are no online standbys or spares to provide a replacement third voter:

https://github.com/canonical/go-dqlite/blob/beebd0121cfa366ebf3cbb9cf9e807af812aa38e/app/roles.go#L162-L164

This PR removes that early return and another one like it, so that we instead continue in the function body and reach the code that demotes the offline node. Thanks to @manadart who reported some unexpected behavior from dqlite role management and caused me to discover this.

### DIFF
--- a/app/app_test.go
+++ b/app/app_test.go
@@ -742,8 +742,8 @@ func TestRolesAdjustment_CantReplaceVoter(t *testing.T) {
 
 	assert.Equal(t, client.Voter, cluster[0].Role)
 	assert.Equal(t, client.Voter, cluster[1].Role)
-	assert.Equal(t, client.Voter, cluster[2].Role)
-	assert.Equal(t, client.StandBy, cluster[3].Role)
+	assert.Equal(t, client.Spare, cluster[2].Role)
+	assert.Equal(t, client.Spare, cluster[3].Role)
 }
 
 // If a stand-by goes offline, another node takes its place.


freeekanayaka:
> Consider a cluster of three dqlite nodes, all of which are voters. If one of these nodes goes offline, we'd like for the leader to eventually demote it to spare during the periodic roles adjustment. This PR fixes a quirk of the `Adjust` function that prevented that demotion from taking place. Specifically, we hit the early return here after noticing that there are no online standbys or spares to provide a replacement third voter:
> 
> https://github.com/canonical/go-dqlite/blob/beebd0121cfa366ebf3cbb9cf9e807af812aa38e/app/roles.go#L162-L164
> 
> This PR removes that early return and another one like it, so that we instead continue in the function body and reach the code that demotes the offline node. Thanks to @manadart who reported some unexpected behavior from dqlite role management and caused me to discover this.
> 
> Signed-off-by: Cole Miller [cole.miller@canonical.com](mailto:cole.miller@canonical.com)

I didn't look at the diff yet, but I'm not entirely sure that demoting that offline voter is going to be helpful in practice.

From an availability point of view, having a cluster of 2 voters both of which are online is no better than having a cluster of 3 voters with only 2 of them being online. That's because in both case the cluster can't tolerate any failure of one of the 2 online voters (i.e. in both cases if one of the 2 online voters goes offline, cluster becomes unavailable). In fact, not demoting the offline voter might be slightly better, since in case it comes back then availability is immediately increased (at that point the cluster can tolerate the loss of 1 online voter), without any time lag for noticing that a spare node is now online again and can be promoted back to voter.

In a situation where voters are being restarted this might be relevant.

freeekanayaka:
> My understanding of the intent of the existing code was that offline nodes should always eventually be demoted to spare; the case of a three-node cluster is just a concrete example where that demotion never happens with the current code. I think the early returns in Adjust are problematic in any case, because that function is structure as a sequence of steps like "if the cluster is imbalanced in X way, make Y change"; with that structure, we only want to return "no changes required" once all the possible imbalances have been checked and found not to exist.

Yes, but I'd say that the 3 nodes only case might be an exception, because of what I described above. There might be other similar exceptions that are not quite dealt with at the moment. However, my point is that I suspect that unconditionally demoting an offline voter might not be the most appropriate choice in some situations, or at least in this specific case I believe that's true.

freeekanayaka:
> So the proposed change doesn't special-case the three-node case, it's a completely generic change that just ensures that our _existing_ step to check for and demote offline nodes always runs.

Yeah, but I unless I'm missing something I would **not** consider the existing behavior of bailing out early buggy or unintended. It actually feels the right thing to do to me, because:

- If you have less online voters than desired **and** you have no node that you could possibly promote to voter in order to improve the situation, then there's no point in demoting existing offline voters, because at least you can hope for them to come back online (e.g. in case of restarts).

- If you have less online stand-bys than desired **and** you have no node that you could possibly promote to stand-by in order to improve the situation, then the same reasoning as above applies.

In other words, I think the two early bail out `return`s [here](https://github.com/canonical/go-dqlite/blob/beebd0121cfa366ebf3cbb9cf9e807af812aa38e/app/roles.go#L163) and [here](https://github.com/canonical/go-dqlite/blob/beebd0121cfa366ebf3cbb9cf9e807af812aa38e/app/roles.go#L198) are a feature and not a bug that needs to be fixed (e.g. by the change in this PR).

Note that the behavior that we want to implement is a bit more subtle than "demote offline nodes no matter what", so I don't think there's anything to fix. Again, unless I'm missing something which makes the current logic sub-optimal.

> If we want to add a special case for a three-node cluster (and I have no strong opinion either way on this) we could make that change separately, either as part of this PR or afterward. But I think if we care about demoting offline nodes at all, we should merge this change, because it applies to clusters with any number of nodes.
> 

I think the confusion comes from the "if we care about demoting offline nodes at all" part: as said, I don't think **unconditionally** demoting offline nodes is the intent of the existing code nor is what we should be aiming for. Under the circumstances I described above, we shouldn't demote offline nodes, and the current code seems to indeed work that way.

> I probably confused the issue by bringing up the three-node case right from the start; sorry about that.

I didn't read the full change when I posted the first comment, I now did and I realize indeed that this is not only about the three-node case. Still, the same reasoning applies, as far as I understand it.

freeekanayaka:
IOW demoting offline voter nodes should only be done if there are enough online voter nodes. Same for stand-bys.

PR #264, opened 2023-07-31
Expose auto-recovery config switch from dqlite

cole-miller:
See https://github.com/canonical/jepsen.dqlite/issues/110

### DIFF
--- a/app/options.go
+++ b/app/options.go
@@ -189,6 +189,20 @@ func WithDiskMode(disk bool) Option {
 	}
 }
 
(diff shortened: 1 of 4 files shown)


app/options.go
```diff
@@ -189,6 +189,12 @@ func WithDiskMode(disk bool) Option {
 	}
 }
 
+func WithAutoRecovery(recovery bool) Option {
```
freeekanayaka, line 192: This needs a docstring, I'm actually surprised that linting didn't fail.

freeekanayaka (approved):

PR #270, opened 2023-09-11
driver: detect EOF on Connect and mark with ErrBadConn.

### LINKED ISSUE #265: Jepsen failure: EOF when calling Open
```
023/08/15 09:26:03.724303 open database: failed to open database: call open (budget 0s): receive: header: EOF
```

We should probably call `driverError` on the wrapped errors

https://github.com/canonical/go-dqlite/blob/f54a1528102399f99af4c1a8ebdbe14db2c68fda/driver/driver.go#L259-L278

MathieuBordere:
This should fix #265 

When calling `app.Open` https://github.com/canonical/go-dqlite/blob/64c24c752b79f72bf12591a53939c9c31c27b96c/app/app.go#L462-L484

the call to `sql.Open` will always return `err == nil` because we don't implement `DriverContext` (see https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/database/sql/sql.go;drc=894d24d617bb72d6e1bed7b143f9f7a0ac16b844;l=813

In the next phase of `app.Open`, `db.PingContext(ctx)` can fail if we've been given a connection that's been closed and returns `EOF` but is not marked with `errBadConn`. (see https://cs.opensource.google/go/go/+/refs/tags/go1.18.1:src/database/sql/sql.go;drc=2580d0e08d5e9f979b943758d3c49877fb2324cb;l=859)

The solution is to also call `driverError` in `connector.Connect` which will mark a connection with `errBadConn` if `EOF` is detected causing the sql package to retry with a new connection. `connector.Connect` is the location where the part of the error message `failed to open database` in #265 comes from.

### DIFF
--- a/driver/driver.go
+++ b/driver/driver.go
@@ -258,7 +258,7 @@ func (c *Connector) Connect(ctx context.Context) (driver.Conn, error) {
 	var err error
 	conn.protocol, err = connector.Connect(ctx)
 	if err != nil {
-		return nil, errors.Wrap(err, "failed to create dqlite connection")
+		return nil, driverError(conn.log, errors.Wrap(err, "failed to create dqlite connection"))
 	}
 
 	conn.request.Init(4096)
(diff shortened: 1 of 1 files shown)


tomponline (approved):
Thanks!

tomponline:
Looks like we have a test failure.

MathieuBordere:
> Looks like we have a test failure.

Can reproduce.

MathieuBordere:
@tomponline So the failure is unrelated to this PR and is being fixed in raft in https://github.com/canonical/raft/pull/477. I will merge this.

PR #276, opened 2023-09-22
OpenTelemetry Driver Tracing

SimonRichardson:
To aid with debugging and development, the following implements the ability to trace requests to the driver. This includes adding the current SQL from conn methods ExecContext or QueryContext, and also adding the stmt ExecContext and QueryContext. Unfortunately, we can't trace every method as the go driver library doesn't allow passing a context to Commit or Rollback.

This is the lightest touch with adding tracing. The tracing code does not require any external dependencies, as this retrieves a Tracer from the passed-in context. As long as the Tracer conforms to the tracing.Tracer interface then a trace can be performed.

There will always be a Tracer if one isn't located on the context.

Potentially, this could supersede the current tracing implementation and remove the need to have a tracing log.

## Example

Currently, the Juju project is starting to implement OpenTelemetry to better understand requests within the system. As the dqlite project is a big part of Juju it would be good if we could see what's happening within the go-dqlite library. I've added a work-in-progress [tempo](https://grafana.com/oss/tempo/) screenshot to show what's available with this PR and what information we might start to gather.

Screenshot from 2023-09-22 22-02-24

### DIFF
--- a/tracing/tracing.go
+++ b/tracing/tracing.go
@@ -8,33 +8,24 @@ const (
 	traceContextKey contextKey = "trace"
 )
 
-// TracerFromContext returns a tracer from the context. If no tracer is found,
-// an empty tracer is returned.
-func TracerFromContext(ctx context.Context) Tracer {
-	value := ctx.Value(traceContextKey)
-	if value == nil {
-		return noopTracer{}
-	}
-	tracer, ok := value.(Tracer)
-	if !ok {
-		return noopTracer{}
-	}
-	return tracer
-}
-
-// WithTracer returns a new context with the given tracer.
+// WithTracer returns a context with the tracer embedded in the context
+// under the context key.
 func WithTracer(ctx context.Context, tracer Tracer) context.Context {
-	if tracer == nil {
-		tracer = noopTracer{}
-	}
 	return context.WithValue(ctx, traceContextKey, tracer)
 }
 
 // Start returns a new context with the given trace.
+// A valid span is always returned, even if the context does not contain a
+// tracer. In that case, the span is a noop span.
 func Start(ctx context.Context, name, query string) (context.Context, Span) {
-	// Tracer is always guaranteed to be returned here. If there is no tracer
-	// available it will return a noop tracer.
-	tracer := TracerFromContext(ctx)
+	value := ctx.Value(traceContextKey)
+	if value == nil {
+		return ctx, noopSpan{}
+	}
+	tracer, ok := value.(Tracer)
+	if !ok {
+		return ctx, noopSpan{}
+	}
 	return tracer.Start(ctx, name, query)
 }
 
@@ -63,13 +54,6 @@ type Span interface {
 	// is called. Therefore, updates to the Span are not allowed after this
 	// method has been called.
 	End()
-}
-
-// noopTracer is a tracer that does nothing.
-type noopTracer struct{}
-
-func (noopTracer) Start(ctx context.Context, name, query string) (context.Context, Span) {
-	return ctx, noopSpan{}
 }
 
 // noopSpan is a span that does nothing.
--- a/tracing/tracing.go
+++ b/tracing/tracing.go
@@ -1,0 +1,82 @@
+package tracing
+
+import "context"
+
+type contextKey string
+
+const (
+	traceContextKey contextKey = "trace"
+)
+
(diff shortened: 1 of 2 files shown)


cole-miller:
Thanks! I'm not familiar with OpenTelemetry but I will take a look at this and see what I can make of it. We really need a better tracing/observability approach for dqlite more broadly, I'm not sure whether OTel is the way to go there, but I do generally agree that the current ad-hoc logging leaves a lot to be desired and we could think about removing it entirely once something better is in place.

SimonRichardson:
So Juju is trying to integrate with the COS ([Canonical Observability Stack](https://discourse.charmhub.io/t/canonical-observability-stack/5132)) framework, which uses [Tempo](https://grafana.com/oss/tempo/) for tracing. Note: This tracing is different from DTrace/strace. They serve different objectives.

We're attempting to provide SREs with a turnkey solution to observe and maintain Juju directly. Selfishly speaking, I want to be able to use otel to give me active traces and understand what may be causing Juju to slow down/or if there are excessive requests in certain parts of the code. Now, I know that it might not tell me exactly, but that's fine. More investigation will probably be required, but if it can help identify issues, then we're in a better place than where Juju is currently.

In addition to this, there is [Polar Signals](https://www.polarsignals.com/) cloud which we're also experimenting with, which will give you the ability to continuously profile. This is analogous to tracing. Profiling tells you what's happening, but generally doesn't support why, which is why we're tackling it from both viewpoints.

SimonRichardson:
Just FYI: it is possible to provide a way of using the new tracing API (start and end) to instrument in the prior way. So we could migrate away from the old way, whilst still offering the new way. I've stuck it up as a prototype on my repo as a potential path forward. https://github.com/SimonRichardson/go-dqlite/pull/1

tracing/tracing.go
```diff
@@ -0,0 +1,82 @@
...
+	return tracer
+}
+
+// WithTracer returns a new context with the given tracer.
+func WithTracer(ctx context.Context, tracer Tracer) context.Context {
+	if tracer == nil {
+		tracer = noopTracer{}
+	}
+	return context.WithValue(ctx, traceContextKey, tracer)
+}
```
cole-miller, line 31: Does this definition need to be in go-dqlite? My understanding is that creating a context with tracing powers is the responsibility of downstream, so it would make more sense to me to just expose the value of `traceContextKey` instead.

tracing/tracing.go
```diff
@@ -0,0 +1,82 @@
...
+
+// Tracer is the interface that all tracers must implement.
+type Tracer interface {
+	// Start creates a span and a context.Context containing the newly-created
+	// span.
+	//
+	// If the context.Context provided in `ctx` contains a Span then the
+	// newly-created Span will be a child of that span, otherwise it will be a
+	// root span.
```
cole-miller, line 48: How does an implementor of Tracer retrieve the "parent" span from the context? Presumably there's another string key that's used?

tracing/tracing.go
SimonRichardson: The rationale behind not exposing the key and exposing the function is that gives you the option of changing the key or how the trace is added to the context. It's generally a [go pattern](https://medium.com/@matryer/context-keys-in-go-5312346a868d) to do it like this for context-additive operations.

I'm not wedded to this pattern if you do not want to expose this function, but instead offer the `TraceContextKey` (as public) then I'm happy to do that as well.

tracing/tracing.go
SimonRichardson: The context that's passed in also contains the parent. If there is another trace started with the threaded context that has a parent id, then it will create a child span of the parent. Not all tracing libraries work this way, but [OTEL](https://opentelemetry.io/) does. We could add another string (or type) for the parent/child relationship, but I'd propose that we cross that bridge when we come to it.

tracing/tracing.go
cole-miller: No, that's fine, we should use the conventional Go style for this.

tracing/tracing.go
cole-miller: I think I understand now, the responsibility for somehow stashing the parent Span inside the context is entirely the implementor's -- I got confused because there was no declared protocol for retrieving the parent, but there doesn't need to be, each implementor can implement their own solution since they control all the relevant code.

cole-miller (approved):

cole-miller:
Thanks!

PR #279, opened 2023-10-17
internal/bindings: Fall back if dqlite_node_set_auto_recovery is missing

cole-miller:
This uses weak linkage to implement a graceful fallback at run time when the dqlite_node_set_auto_recovery symbol is not provided by libdqlite. With this merged, we can cut a release of go-dqlite without increasing the minimum supported libdqlite version.


cole-miller:
@MathieuBordere -- I did experiment with doing this without the weak symbol trick, but that seems to be a necessary ingredient. Otherwise the resulting undefined symbol error from the dynamic loader can't be handled.

PR #280, opened 2023-10-19
.github/workflows: run tests on go 1.21.x

MathieuBordere:


cole-miller:
Does it make sense to remove the earliest Go version from the matrix at the same time? I don't know what our Go support policy is.

MathieuBordere:
> Does it make sense to remove the earliest Go version from the matrix at the same time? I don't know what our Go support policy is.

We'd have to check what the go version is in Focal, I would support the oldest default go version in Focal.

PR #286, opened 2024-02-03
driver_test: test interrupting query yielding rows.

MathieuBordere:
Actually test sending an interrupt.


MathieuBordere:
Merging this one.

PR #292, opened 2024-03-17
Attempt to connect to leader via 10 servers simultaneously

### LINKED ISSUE #134: Parallellize connectAttemptAll

cnnrznn:
This PR implements the solution discussed in my [first PR](https://github.com/canonical/go-dqlite/pull/290)

1. Spawn connection attempts in goroutines
2. Goroutines
  a. Grab semaphore, defer release
  b. Check if a connection has been found, exit
  c. Execute original logic, write protocol output to channel

Resolves #134

### DIFF
--- a/internal/protocol/connector.go
+++ b/internal/protocol/connector.go
@@ -7,6 +7,7 @@ import (
 	"io"
 	"net"
 	"sort"
+	"sync"
 	"time"
 
 	"github.com/Rican7/retry"
@@ -14,6 +15,7 @@ import (
 	"github.com/Rican7/retry/strategy"
 	"github.com/canonical/go-dqlite/logging"
 	"github.com/pkg/errors"
+	"golang.org/x/sync/semaphore"
 )
 
 // DialFunc is a function that can be used to establish a network connection.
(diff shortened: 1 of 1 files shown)


internal/protocol/connector.go
```diff
@@ -126,56 +128,98 @@ func (c *Connector) connectAttemptAll(ctx context.Context, log logging.Func) (*P
...
-		log := func(l logging.Level, format string, a ...interface{}) {
-			format = fmt.Sprintf("server %s: ", server.Address) + format
-			log(l, format, a...)
-		}
+		go func(server NodeInfo, pc chan<- *Protocol) {
+			defer wg.Done()
 
-		ctx, cancel := context.WithTimeout(ctx, c.config.AttemptTimeout)
-		defer cancel()
+			sem.Acquire(ctx, 1)
```
SimonRichardson, line 151: This can return an error. We should handle that case, and if you release too many times, it will cause a panic. We should be careful in that scenario.

cnnrznn: Updated, following the [example](https://pkg.go.dev/golang.org/x/sync/semaphore#example-package-WorkerPool)

cnnrznn: There is still [the case](https://cs.opensource.google/go/x/sync/+/refs/tags/v0.6.0:semaphore/semaphore.go;drc=43a5402ce75a95522677f77c619865d66b8c57ab;l=109) where the semaphore can be acquired after cancellation. To sacrifice performance but improve readability we could also eliminate (new) lines 156-158

SimonRichardson: We do exactly the same in juju 😉

internal/protocol/connector.go
```diff
@@ -126,56 +128,100 @@ func (c *Connector) connectAttemptAll(ctx context.Context, log logging.Func) (*P
...
+				// the leader, try with the next target.
+				log(logging.Warn, "reported leader server is not the leader")
+				return
+			}
+			log(logging.Debug, "connected")
+			pc <- protocol
+		}(server, protocolChan)
+	}
+
+	// Read from protocol chan, cancel context\
```
cole-miller, line 213: typo

internal/protocol/connector.go
```diff
@@ -126,56 +128,100 @@ func (c *Connector) connectAttemptAll(ctx context.Context, log logging.Func) (*P
...
+		if protocol != nil {
+			pc.Close()
 			continue
 		}
-		log(logging.Debug, "connected")
+		protocol = pc
+		cancel()
+	}
+
+	if protocol != nil {
```
cole-miller, line 224: I feel this would be a bit clearer as follows:

```go
protocol, ok := <-protocolChan
if !ok {
    return nil, ErrNoAvailableLeader
}
cancel()
for extra := range protocolChan {
    extra.Close()
}
return protocol, nil
```

cole-miller (changes requested):
Thanks again for working on this and sorry for the delay! Only minor comments.

internal/protocol/connector.go
cnnrznn: I like it

cole-miller:
I don't think it's necessary!

cole-miller (approved):

cnnrznn:
:rocket: woohoo! :rocket:

PR #296, opened 2024-05-09
make database concurrent

SaumyaSachdev:


driver/driver.go
```diff
@@ -384,38 +384,51 @@ func (c *Conn) Prepare(query string) (driver.Stmt, error) {
...
+        if err != nil && c.tracing != client.LogNone {
 
-	var result protocol.Result
-	result, err = protocol.DecodeResult(&c.response)
-	if err != nil {
-		return nil, driverError(c.log, err)
-	}
+			duration := 20 * time.Microsecond
+			time.Sleep(duration)
+            c.log(c.tracing, "error in exec context %v", err)
```
SimonRichardson, line 419: @SaumyaSachdev question are you attempting to retry on failure?

Currently, the likes of LXD and Juju are retrying at a higher level (see LXD for an example):

https://github.com/canonical/lxd/blob/d5b1d00f5651a6b46541e6787d4f9a6ed494f179/lxd/db/query/retry.go#L25-L58

I would like to move this into the driver, but it should be configurable if it was placed here.


Note: the title to the PR is incorrect for what is happening.

SaumyaSachdev:
Apologies, this PR was opened in error.

PR #302, opened 2024-07-05
Check if leader was found in between network requests

masnax:
We send a request to multiple peers concurrently to ask if they are the leader. They themselves may be the leader, or they may report who they think is the leader. We then try to reach out to whoever they think is the leader.

It is possible that we are are already talking to the leader concurrently with asking a different node who the leader is, so in such cases we don't need to send further network requests and can return early by checking if we have already found the leader in another goroutine.

### DIFF
--- a/internal/protocol/connector.go
+++ b/internal/protocol/connector.go
@@ -143,6 +143,16 @@ func (c *Connector) connectAttemptAll(ctx context.Context, log logging.Func) (*P
 		close(protocolChan)
 	}()
 
+	var found bool
+	foundLock := sync.RWMutex{}
+	checkFound := func() bool {
+		foundLock.RLock()
+		defer foundLock.RUnlock()
+		localFound := found
+
+		return localFound
+	}
+
 	// Make an attempt for each address until we find the leader.
 	for _, server := range servers {
 		go func(server NodeInfo, pc chan<- *Protocol) {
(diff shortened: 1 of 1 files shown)


cole-miller:
We already have some logic along these lines---the investigating goroutines share a `Context`:

https://github.com/canonical/go-dqlite/blob/8145904735576ae350c1649426eab29da9f2c0e1/internal/protocol/connector.go#L131

Which is passed into `connectAttemptOne`:

https://github.com/canonical/go-dqlite/blob/8145904735576ae350c1649426eab29da9f2c0e1/internal/protocol/connector.go#L168

And this context is cancelled as soon as we get a connection to the leader:

https://github.com/canonical/go-dqlite/blob/8145904735576ae350c1649426eab29da9f2c0e1/internal/protocol/connector.go#L214-L219

So this change might be redundant, if I'm not missing something.

masnax:
> So this change might be redundant, if I'm not missing something.

Ah, looks like this one is a bug on our end then!

I was doing 2 things incorrectly here:
* I was counting the calls to the DialFunc instead of the calls that make it through to the other side. With the context, the cancellation won't happen until much later.
* microcluster is not correctly passing the context through as it uses [tls.DialWithDialer](https://pkg.go.dev/crypto/tls#DialWithDialer) which internally uses `context.Background()`. Hence why I saw an improvement there, but now checking with `dqlite-demo` I don't see the same improvement.

PR #303, opened 2024-07-05
Configure maximum concurrent leader probes

masnax:
Recently, `go-dqlite` was updated in #292 to concurrently open up to 10 requests to other cluster members when probing for leadership. The goal of this was to prevent go-dqlite from getting hung up on some unresponsive nodes. 

One issue with this is that it is a bit too aggressive, each attempt to fetch the leader will execute effectively `2n - 1` network requests where `n` is the number of cluster members. #302 will make this return early if the leader is found, but it would also be great if a project could configure this value to grow and shrink based on the current health of the cluster. 

Effectively, if the cluster is healthy, the `2n -1` requests are redundant because we generally only need a maximum of 2. (One to some node and another to who that node reports is the leader). 

So this PR adds the `WithConcurrentLeaderConns` option to each of `client`, `driver`, and `app`. The default value will remain 10 as before.

In particular for `app` and `driver`, the value is actually recorded as a pointer so that a project can dynamically grow and shrink the maximum concurrent connections that will be maintained when probing for the leader during the role adjustment interval. 

The idea here is that the hook from #301 can be used to check the deviation in node response times after each role adjustment, and then increment or decrement the maximum connections that the cluster can tolerate.

### DIFF
--- a/internal/protocol/connector.go
+++ b/internal/protocol/connector.go
@@ -18,6 +18,9 @@ import (
 	"golang.org/x/sync/semaphore"
 )
 
+// MaxConcurrentLeaderConns is the default maximum number of concurrent requests to other cluster members to probe for leadership.
+const MaxConcurrentLeaderConns int64 = 10
+
 // DialFunc is a function that can be used to establish a network connection.
 type DialFunc func(context.Context, string) (net.Conn, error)
 
@@ -51,6 +54,10 @@ func NewConnector(id uint64, store NodeStore, config Config, log logging.Func) *
 
 	if config.BackoffCap == 0 {
 		config.BackoffCap = time.Second
+	}
+
+	if config.ConcurrentLeaderConns == 0 {
+		config.ConcurrentLeaderConns = MaxConcurrentLeaderConns
 	}
 
 	connector := &Connector{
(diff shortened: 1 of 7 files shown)


cole-miller:
Making the concurrency limit configurable seems reasonable to me; I'd also be totally fine lowering the default from `10` if that works better for LXD.

internal/protocol/connector.go
```diff
@@ -53,6 +53,10 @@ func NewConnector(id uint64, store NodeStore, config Config, log logging.Func) *
 		config.BackoffCap = time.Second
 	}
 
+	if config.ConcurrentLeaderConns == 0 {
+		config.ConcurrentLeaderConns = 10
+	}
```
cole-miller, line 58: Can this be avoided by having `Driver.concurrentLeaderConns` default to MaxConcurrentLeaderConns?

cole-miller (approved):
Only one minor comment, otherwise this looks good!

internal/protocol/connector.go
masnax: Yes, though I had to move the constant to the `protocol` package to avoid an import cycle.

PR #304, opened 2024-07-10
Use range loop instead of endless loop

marco6:
This PR uses range loop instead of infinite loop + recv from channel as it is difficult during application tear-down to synchronize between application shutdown and listener closing. If for some reason the `acceptChan` is closed while the goroutine is still waiting, the received connection is `nil` and makes the code panic.

Using range loop over the channel solves the problem, makes the code more robust and (arguably) more idiomatic.


cole-miller (approved):
Thanks, good catch!

PR #306, opened 2024-07-11
Avoid context cancellation during net.Dial call in connector

### LINKED ISSUE #305: EOF errors on connect when using external connection with `http.Serve`
As a result of #302, I realized microcluster was not properly handling the context it received from `go-dqlite` when the dial func is called.

However, [upon properly propagating this context](https://github.com/canonical/microcluster/pull/166), I began to notice lots of these errors constantly on each heartbeat:
```
2024/07/10 15:57:18 http: TLS handshake error from 127.0.0.1:60120: EOF
2024/07/10 15:57:26 http: TLS handshake error from 127.0.0.1:48016: read tcp 127.0.0.1:9000->127.0.0.1:48016: read: connection reset by peer
```

`http.Serve` logs TLS handshake errors on line 1935 here: https://go.dev/src/net/http/server.go

This looks to be connected to [this cancel() call](https://github.com/canonical/go-dqlite/blob/95f620cdb3f17d65af0eb21a5849cdd210028000/internal/protocol/connector.go#L219) which cancels all active connections using the dial func.  

The errors are too generic to just ignore, so I was thinking if it would be cleaner to simply leave the connections alive and use a mutex to avoid starting new connections when the leader is found to avoid this noise?

cole-miller:
Closes #305.

### DIFF
--- a/internal/protocol/connector.go
+++ b/internal/protocol/connector.go
@@ -140,14 +140,14 @@ func (c *Connector) connectAttemptAll(ctx context.Context, log logging.Func) (*P
 
 	sem := semaphore.NewWeighted(c.config.ConcurrentLeaderConns)
 
-	protocolChan := make(chan *Protocol)
+	protocolCh := make(chan *Protocol)
 
 	wg := &sync.WaitGroup{}
 	wg.Add(len(servers))
 
 	go func() {
 		wg.Wait()
-		close(protocolChan)
+		close(protocolCh)
 	}()
 
 	// Make an attempt for each address until we find the leader.
@@ -214,18 +214,18 @@ func (c *Connector) connectAttemptAll(ctx context.Context, log logging.Func) (*P
 			}
 			log(logging.Debug, "connected")
 			pc <- protocol
-		}(server, protocolChan)
+		}(server, protocolCh)
 	}
 
 	// Read from protocol chan, cancel context
-	protocol, ok := <-protocolChan
+	protocol, ok := <-protocolCh
 	if !ok {
 		return nil, ErrNoAvailableLeader
 	}
 
 	cancel()
 
-	for extra := range protocolChan {
+	for extra := range protocolCh {
 		extra.Close()
 	}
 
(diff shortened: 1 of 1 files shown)


cole-miller:
cc @masnax

masnax:
@cole-miller Confirmed this eliminates the log spam with `http.Serve`, thanks!

internal/protocol/connector.go
```diff
@@ -214,18 +220,18 @@ func (c *Connector) connectAttemptAll(ctx context.Context, log logging.Func) (*P
...
 	if !ok {
 		return nil, ErrNoAvailableLeader
 	}
 
 	cancel()
 
-	for extra := range protocolChan {
+	for extra := range protocolCh {
 		extra.Close()
 	}
```
masnax, line 236: Just a question, with these changes, since we are not cancelling the dial context, does that mean `protocolCh` might still end up blocking this function from returning if a particular node is very slow and happens to have a request in flight?

If so, then I wonder if this loop should be in a goroutine, so that the leader is returned immediately before the remaining connections are closed? But that might cause spikes if there's repeated attempts to fetch the leader back-to-back.

Alternatively there could be another goroutine situated right after the call to `cancel()` which sleeps for some (configurable?) abort limit and then cancels the `origCtx` to potentially unblock `protocolCh`.  At least in my test setup about 100ms is fine enough to get past the TLS handshake, but that's not a very real-world test.

tomponline (approved):

internal/protocol/connector.go
cole-miller: > Just a question, with these changes, since we are not cancelling the dial context, does that mean protocolCh might still end up blocking this function from returning if a particular node is very slow and happens to have a request in flight?

Yep, that's true! I feel like maybe this is an argument for keeping a pool of open connections to the other nodes in the cluster instead of creating a bunch of new ones every time and then throwing them away. This requires dealing with membership changes, but we have a natural place to update the pool, namely the node store refresh hook. Hopefully this would also deal with the excessive CPU usage issues. What do you think?

PR #307, opened 2024-07-15
driver: Expose leadership-related error codes

cole-miller:
These are stable and potentially useful for clients.


tomponline (commented):
sounds useful thanks!

PR #312, opened 2024-08-21
Add a wrapper for dqlite_node_describe_last_entry

### LINKED ISSUE #299: Provide API for finding the most recent segment index
As noted in [this doc](https://github.com/canonical/go-dqlite/blob/master/docs/restore-db.md) (see section A.3.note), some kind of API for determining the highest segment index would make it possible to remove ambiguity when we refer in docs to "the most up to date cluster member". I don't have a super thorough understanding of dqlite/raft internals here but I'm assuming (based on the linked doc) that the highest-index closed segment file is not the most reliable way of determining this (as is [done in LXD](https://github.com/canonical/lxd/blob/098b4952c322ce67c4bb209118823fab6ca51bd0/lxd/db/db.go#L450)).

Happy to open this in the dqlite repo instead if here isn't the right place.

cole-miller:
Closes #299. CI will fail until the dqlite PR merges, and we'll need to either bump the minimum supported libdqlite version or add a shim to support older dqlite versions that lack this function.

### DIFF
--- a/node.go
+++ b/node.go
@@ -239,20 +239,16 @@ func ReconfigureMembership(dir string, cluster []NodeInfo) error {
 // loss and inconsistency:
 //
 // 1. Make sure no dqlite node in the cluster is running.
-//
 // 2. Identify all dqlite nodes that have survived and that you want to be part
 //    of the recovered cluster. Call this the "new member list".
-//
-// 3. From the nodes in the new member list, find the one with the most
-//    up-to-date raft term and log. Call this the "template node".
-//
+// 3. Call ReadLastEntryInfo on each node in the member list, and find which
+//    node has the most recent entry according to LastEntryInfo.Before. Call this
+//    the "template node".
 // 4. Invoke ReconfigureMembershipExt exactly one time, on the template node.
 //    The arguments are the data directory of the template node and the new
 //    member list.
-//
 // 5. Copy the data directory of the template node to all other nodes in the
 //    new member list, replacing their previous data directories.
-//
 // 6. Restart all nodes in the new member list.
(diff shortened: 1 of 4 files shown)


cole-miller:
This needs the PPA to be updated to pass `--enable-build-raft`, see https://github.com/canonical/dqlite-ppa/pull/7

cole-miller:
This remains blocked due to issues with the dqlite PPA not getting updated, see https://github.com/canonical/dqlite/pull/698.

node.go
```diff
@@ -248,6 +248,37 @@ func ReconfigureMembershipExt(dir string, cluster []NodeInfo) error {
...
+	Term, Index uint64
+}
+
+// ReadLastEntryInfo reads information about the last entry in the raft
+// persistent log from a node's data directory.
+//
+// This is a non-destructive operation, but is not read-only, since it has the
+// side effect of renaming raft open segment files to closed segment files.
+func ReadLastEntryInfo(dir string) (LastEntryInfo, error) {
+	server, err := bindings.NewNode(context.Background(), 1, "1", dir)
```
letFunny, line 267: I guess this is creating a new with a fake number and address, hence the 1 and "1" is not important. My question is, can this id clash with some existing node?

letFunny (approved):
Pending the ongoing discussion on MM and my comment here, LGTM!

node.go
cole-miller: Indeed the ID and address are fake, and they can alias with the ID/address of a node that's present in the cluster configuration. However, it doesn't matter because dqlite_node_create doesn't take any action based on the provided ID and address, it just stores them away for dqlite_node_start to use later.

Issue #313, opened 2024-08-21
REPL should output column names


SimonRichardson:
When using the REPL from the dqlite-tools the REPL only prints the data, it would be much more useful to return the column names in the output.

PR #314, opened 2024-08-21
feat: tab writer

SimonRichardson:
Without column headers, it's non-obvious what each column value is without referring to the original schema. To remove the ambiguity and make it easier to read the output this adds a tabwriter from the standard library.

The output is a lot cleaner IMO, but you do lose the separator lines. We could add it back using an external package, but I'm unsure it's worth it.

Fixes: #313

Example output becomes:

```
dqlite> SELECT * FROM charm_origin
charm_uuid                              source_id       id      revision        version
a22ff9d2-3da4-4653-8e00-ea85aa045546    1               <nil>   9               <nil>
03a7ab56-9522-4e46-8f5d-79af83fc762e    1               <nil>   10
```

### DIFF
--- a/internal/shell/shell.go
+++ b/internal/shell/shell.go
@@ -11,6 +11,7 @@ import (
 	"path/filepath"
 	"strconv"
 	"strings"
+	"text/tabwriter"
 
 	"github.com/canonical/go-dqlite"
 	"github.com/canonical/go-dqlite/client"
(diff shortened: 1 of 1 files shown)


cole-miller (approved):
Thanks, this is a nice change!

cole-miller:
@neoaggelos @tomponline Just want to check that MicroK8s and LXD don't parse the output of the dqlite client CLI in a way that depends on `|` separators before merging this.

neoaggelos:
No, we are good on the Kubernetes side. Cannot tell if something downstream might be doing this

tomponline:
@cole-miller im not sure, we have the `lxd sql global` command which allows executing arbitrary commands but its not interactive, so potentially doesn't use this.

cole-miller:
@tomponline Indeed, looks like `lxd sql global` doesn't go through our shell since the output format is quite different. I think we're good to merge this then.

PR #315, opened 2024-08-23
feat(connector): honor context cancelation

letFunny:
Honor context cancelation and primitive version of error reporting. We need to revamp the error handling and reporting for us to be able to provide better ones, until then we will follow the convention in the code.

As part of this PR, I did some changes to the *Message functions to make them public to properly test the code. Ideally they will live in their own package with proper encapsulation but that can come at a later PR.


internal/utils/afterfunc.go
```diff
@@ -0,0 +1,36 @@
...
+				close(done)
+				f()
+			}
+		case <-done:
+		}
+	}()
+
+	return func() bool {
+		if atomic.CompareAndSwapInt32(&run, 0, 1) {
+			close(done)
```
just-now, line 31: is it expected that under certain circumstances the `done` channel will be `closed()` twice (lines 22 and 31)?

just-now: ah. now i see, there's an atomic to guarantee it.

internal/protocol/connector.go
```diff
@@ -236,17 +238,30 @@ func Handshake(ctx context.Context, conn net.Conn, version uint64) (*Protocol, e
...
+			// Cancel read and writes by setting deadline in the past.
+			conn.SetDeadline(time.Unix(1, 0))
+		}
+	})
+	defer stop()
 
 	// Perform the protocol handshake.
 	n, err := conn.Write(protocol)
 	if err != nil {
+		if canceled && errors.Is(err, os.ErrDeadlineExceeded) {
```
cole-miller: `canceled` is being written from one goroutine and read from another, is there a race?

cole-miller: Also why do we have a separate `if` for this case when it's handled identically to the one just below?

internal/protocol/connector_test.go
```diff
@@ -93,8 +94,8 @@ func TestConnector_EmptyNodeStore(t *testing.T) {
 	check([]string{})
 }
 
-// Connection failed because the context was canceled.
-func TestConnector_ContextCanceled(t *testing.T) {
+// Connection failed because the context expired.
+func TestConnector_ContextExpired(t *testing.T) {
```
cole-miller, line 98: Good catch on the name

internal/protocol/connector_test.go
```diff
@@ -111,6 +112,44 @@ func TestConnector_ContextCanceled(t *testing.T) {
 	})
 }
 
+// Connection failed because the context expired.
```
cole-miller, line 116: s/expired/canceled?

internal/protocol/connector_test.go
```diff
@@ -107,9 +107,10 @@ func TestConnector_ContextExpired(t *testing.T) {
 	_, err := connector.Connect(ctx)
 	assert.Equal(t, protocol.ErrNoAvailableLeader, err)
 
-	check([]string{
-		"WARN: attempt 1: server 1.2.3.4:666: dial: dial tcp 1.2.3.4:666: i/o timeout",
-	})
+	// TODO: use check when there is an interface for regexp.
+	/*check([]string{
+		"WARN: attempt 1: server 1.2.3.4:666: dial: dial tcp 1.2.3.4:666: (i/o timeout|deadline exceeded)",
+	})*/
```
cole-miller, line 113: I am confused why the error message changes from version to version because [this commit](https://github.com/golang/go/commit/d422f54619b5b6e6301eaa3e9f22cfa7b65063c8) makes an effort to preserve the error message when introducing `os.ErrDeadlineExceeded`, do you know what's going on with this?

internal/protocol/connector.go
```diff
@@ -249,17 +252,33 @@ func Handshake(ctx context.Context, conn net.Conn, version uint64) (*Protocol, e
...
+		}
+	})
+	defer stop()
 
 	// Perform the protocol handshake.
+	// TODO: this should be a message and use same code for writing and reading
+	// to the network.
 	n, err := conn.Write(protocol)
 	if err != nil {
+		nerr, ok := errors.Cause(err).(net.Error)
```
cole-miller, line 271: Naive Go question, do we know for sure that we can use `errors.Cause` here i.e. that the error from `conn.Write` (and in other places) conforms to the expected interface? Is this something stable that the standard library can't break?

internal/utils/afterfunc.go
```diff
@@ -0,0 +1,36 @@
...
+// +build !go1.21
+
+package utils
+
+import (
+	"context"
+	"sync/atomic"
+)
+
+func AfterFunc(ctx context.Context, f func()) (stop func() bool) {
```
cole-miller, line 11: This implementation seems reasonable to me but can you elaborate on how you got to it? If I go to the source code for AfterFunc in the Go repo I see something that looks different and I want to make sure I understand why.

cole-miller (changes requested):
Thank you Alberto! Bunch of comments but they are mostly just me asking questions.

internal/protocol/connector.go
letFunny: Changed the error message to say canceled, this was more for prepping for the next change but it makes sense to just add it now. Thanks for spotting it!

internal/protocol/connector.go
letFunny:
```
// Cause returns the underlying cause of the error, if possible.
// An error value has a cause if it implements the following
// interface:
//
//     type causer interface {
//            Cause() error
//     }
//
// If the error does not implement Cause, the original error will
// be returned. If the error is nil, nil will be returned without further
// investigation.
```

I would say it is safe to use because even if it is nil the type assertion should not work.

internal/protocol/connector_test.go
letFunny: I think it is because of https://github.com/golang/go/commit/d422f54619b5b6e6301eaa3e9f22cfa7b65063c8#diff-898358dce6794a96903071740a43b22d1985d2dc4163e7fbbb65df3b816736d6L81-L86

internal/utils/afterfunc.go
letFunny: Tests added, now we can verify that it conforms to the official requirements per the context.AfterFunc documentation.

cole-miller: Thanks, the tests are great to have!

letFunny:
Closing now that there has been an attempt to refactor the connector / client classes.

PR #318, opened 2024-09-05
Propagate dqlite_node_recover_ext errors

petrutlucian94:
This commit ensures that the dqlite recovery error messages are propagated by go-dqlite.


cole-miller (approved):
Thanks!

PR #333, opened 2024-10-02
Client support for non-leader connections

cole-miller:
This is the go-dqlite counterpart of canonical/dqlite#709, tests are expected to fail since that hasn't merged. We expose the new functionality with a query parameter in the filename passed to `OpenConnector`. The usage is

```
connector := dqlitedriver.OpenConnector("file:my_db_name?dqlite_direct=TARGET")
conn, err := connector.Connect(ctx)
// conn is a database/sql/driver.Conn, needs some type-asserting to use
```

This gives you a usable connection to the node with address `TARGET`, which need not be the leader. You can prepare statements and run queries.


sabaini:
Would this API make it possible to force leader reads for specific queries?

cole-miller:
@sabaini With the current state of the PR you'd have to store two DB connections, a "normal" one that talks to the currrent leader and a special one created using the new API that talks to a specific (possibly non-leader) node. I've implemented it this way because go-dqlite is built on a model where one DB connection = one network connection to a dqlite server; moving away from that would introduce a lot more complexity, especially with the `database/sql` connection pool in the mix.

marco6:
dqlite does not support at the moment the proper semantics for reads on followers. Closing.

PR #338, opened 2024-10-08
Recovery support for App

cole-miller:
WIP, adding an `App`-aware way to do recovery, cc @MggMuggins

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -1,13 +1,18 @@
 package app
 
 import (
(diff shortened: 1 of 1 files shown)


MggMuggins:
I don't want to use a formal review here as I haven't inspected this at that level of detail, but from a high level I've given this a think and it looks workable. Thanks so much for keeping this in mind!

PR #340, opened 2024-10-22
Release v1.99.99 and retract v1 series

cole-miller:


cole-miller:
It didn't occur to me to check this, but `retract` didn't exist until Go 1.16, so we can't use it to point people away from the v1 branch. Therefore I'm just going to tag the final v1 release without this and hopefully the release notes will be enough to push people in the right direction.

tomponline:
> It didn't occur to me to check this, but `retract` didn't exist until Go 1.16, so we can't use it to point people away from the v1 branch. Therefore I'm just going to tag the final v1 release without this and hopefully the release notes will be enough to push people in the right direction.

Could you bump min version to 1.16 first?

cole-miller:
@tomponline Was hesitant to do this initially but realized it's only needed to increase the minimum Go version on master, not for v2 or v3, so going that route, thanks!

PR #342, opened 2024-10-22
Bump minimum libdqlite to v1.17 for LTS

cole-miller:


cole-miller:
Needs LTS libdqlite packages for the PPA to be set up first, working on it

PR #361, opened 2025-02-10
feat: new snapshot trailing strategy parameter

marco6:
This PR adds the capability to set the new strategy parameter for trailing entries computation.

In case the API is not there (because of an old library linked) it will still accept settings with "STATIC" as that's the old behaviour. Trying to set "DYNAMIC" behaviour on old libraries will result in an error.


just-now (approved):
lgtm

louiseschmidtgen (approved):
LGTM, thanks Marco!

PR #369, opened 2025-03-06
feat: add bindings for busy timeout parameter

marco6:
This PR adds bindings for the new (v1.18.2) server-side busy timeout.

### DIFF
--- a/internal/bindings/server.go
+++ b/internal/bindings/server.go
@@ -146,12 +146,12 @@ func init() {
 
 // NewNode creates a new Node instance.
 func NewNode(ctx context.Context, id uint64, address string, dir string) (*Node, error) {
-	requiredVersion := dqliteMajorVersion*100 + dqliteMinorVersion
 	// Remove the patch version, as patch versions should be compatible.
-	runtimeVersion := int(C.dqlite_version_number()) / 100
-	if requiredVersion > runtimeVersion {
+	if DqliteVersion < DqliteMinSupportedVersion {
+		requiredMajor, requiredMinor, _ := versionComponents(DqliteMinSupportedVersion)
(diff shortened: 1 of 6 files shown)


marco6:
Depends on canonical/dqlite#736.

marco6:
Ready as canonical/dqlite#736 was merged.

internal/bindings/server.go
```diff
@@ -91,6 +91,20 @@ static int setSnapshotParameters(dqlite_node *n, unsigned snapshot_threshold, un
...
+__attribute__((weak))
+int dqlite_node_set_busy_timeout(dqlite_node *n, unsigned msecs);
+
+static int setBusyTimeout(dqlite_node *n, unsigned msecs) {
+	if (dqlite_node_set_busy_timeout) {
+		return dqlite_node_set_busy_timeout(n, msecs);
+	}
+	if (msecs == 0) {
+		return 0;
+	}
```
letFunny, line 104: I think this branch is never taken, we are already checking for the default in `node.go`.

letFunny (commented):
Thanks for this Marco, I have two high level comments:
* I think the wording in the comments should be precise just like what you wrote. However, I also think we should have a bit more prose for more unfamiliar users that will guide them more in figuring out what each option is doing.
* Do we need tests for this? I wouldn't like to cause a panic in prod because a symbol is not defined in the old dqlite and breaking everyone that upgrades.

internal/bindings/server.go
marco6: I decided to keep it as a safeguard in the future as I don't like to rely on upper layer for this. It also makes the testing somewhat easier.

letFunny (approved):

PR #372, opened 2025-03-17
add `exit` command to quit dqlite shell

legionxiong:


marco6:
Hi! Thanks for this PR. Could you rebase on v3 now that the cla-check and the pipeline is (hopefully) fixed?

Also, I would prefer to be consistent in this case with the `sqlite3` interface where `.quit` is used instead of `.exit` (there exists a `.exit`, but it is used to return an error code).

PR #385, opened 2025-07-10
feat: add support for 1.17 back

marco6:


letFunny:
@marco6 I see that TestIntegration_ExecBindError is failing for some versions.

marco6:
> I see that TestIntegration_ExecBindError is failing for some versions.

@letFunny it is failing for all 1.18.x versions and will keep doing that as the `EXEC_SQL` command is broken by design. If you remember, we had 2 choices (which the current test clearly is not highlighting):

 - support binding only on the first statement (which was the case for 1.17) so `db.Exec("INSERT INTO a VALUES (?); INSERT INTO b VALUES (?)", 1, 2)` fails (and it is clearly not tested as it wouldn't have looked great!) but on the other side `db.Exec("INSERT INTO a VALUES (?)", 1, 2)`  is checked correctly and fails
 - supporting binding on multiple statements (which is the case for 1.18) so the exact reverse. It would be a bit weird to return "error".
 
 There is a "fix" that uses a prepared statement in #383.

just-now (approved):
Looks good.

letFunny:
@marco6 so you are fixing the tests as part of another PR, then it's okay for me. I just don't want tests always failing.

internal/bindings/server.go
```diff
@@ -72,13 +72,20 @@ static int setAutoRecovery(dqlite_node *t, bool on) {
...
+#endif
+
 __attribute__((weak))
 int dqlite_node_set_snapshot_params_v2(dqlite_node *n, unsigned snapshot_threshold, unsigned snapshot_trailing, int trailing_strategy);
 
 static int setSnapshotParameters(dqlite_node *n, unsigned snapshot_threshold, unsigned snapshot_trailing, int trailing_strategy) {
 	if (dqlite_node_set_snapshot_params_v2) {
 		return dqlite_node_set_snapshot_params_v2(n, snapshot_threshold, snapshot_trailing, trailing_strategy);
-	} else if (trailing_strategy == DQLITE_SNAPSHOT_TRAILING_STATIC) {
+	} else if (trailing_strategy == 0) {
```
letFunny: Why use 0 if by the macro above this constants are always defined? I am probably missing somehting

letFunny (approved):
Just one question and LGTM!

internal/bindings/server.go
marco6: Nicely spotted, that was coming from a previous iteration. I'll go back to the named value.

PR #387, opened 2025-07-11
chore: add security scan

marco6:


.github/workflows/security.yml
```diff
@@ -0,0 +1,111 @@
...
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+
+      - name: Install gosec
+        run: go install github.com/securego/gosec/v2/cmd/gosec@latest
```
simondeziel, line 69: It's the first time I see this `gosec` tool but it seems they make it easy to integrate with GH actions, see https://github.com/securego/gosec?tab=readme-ov-file#github-action

PR #389, opened 2025-08-06
feat: support removal of disk mode

marco6:
This PR deprecates `WithDiskMode` options and supports any new version of `dqlite` (likely `1.18.3`) that doesn't have disk mode.

### DIFF
--- a/app/options.go
+++ b/app/options.go
@@ -200,10 +200,8 @@ func WithSnapshotParams(params dqlite.SnapshotParams) Option {
 }
 
 // WithDiskMode enables or disables disk-mode.
-// WARNING: This is experimental API, use with caution
-// and prepare for data loss.
-// UNSTABLE: Behavior can change in future.
-// NOT RECOMMENDED for production use-cases, use at own risk.
+// DEPRECATED: this API will always fail on dqlite 1.18.3+ as support for
+// disk mode has been dropped.
 func WithDiskMode(disk bool) Option {
 	return func(options *options) {
 		options.DiskMode = disk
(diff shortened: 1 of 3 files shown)


marco6:
See canonical/dqlite#801.

app/options.go
```diff
@@ -200,10 +200,8 @@ func WithSnapshotParams(params dqlite.SnapshotParams) Option {
...
 
 // WithDiskMode enables or disables disk-mode.
-// WARNING: This is experimental API, use with caution
-// and prepare for data loss.
-// UNSTABLE: Behavior can change in future.
-// NOT RECOMMENDED for production use-cases, use at own risk.
+// DEPRECATED: this API will always fail on dqlite 1.18.3+ as support for
+// disk mode has been dropped.
 func WithDiskMode(disk bool) Option {
 	return func(options *options) {
```
letFunny, line 206: Why not make it fail here directly?

node.go
```diff
@@ -84,10 +84,8 @@ func WithSnapshotParams(params SnapshotParams) Option {
 }
 
 // WithDiskMode enables dqlite disk-mode on the node.
-// WARNING: This is experimental API, use with caution
-// and prepare for data loss.
-// UNSTABLE: Behavior can change in future.
-// NOT RECOMMENDED for production use-cases, use at own risk.
+// DEPRECATED: this API will always fail on dqlite 1.18.3+ as support for
+// disk mode has been dropped.
 func WithDiskMode(disk bool) Option {
```
letFunny, line 89: Same comment, let's make it fail here directly.

internal/bindings/server.go
```diff
@@ -215,7 +226,7 @@ func (s *Node) SetFailureDomain(code uint64) error {
 
 func (s *Node) EnableDiskMode() error {
 	server := (*C.dqlite_node)(unsafe.Pointer(s.node))
```
letFunny, line 250: This can also fail directly.

app/options.go
marco6: It is the job of this Go wrapper to support libdqlite in Go. This is true for older version as well. So, if someone is using `v1.17.x` or `v1.18.{0-3}` it should still be possible to enable disk mode.

Also, this method does not return an error. Are you suggesting a panic?

node.go
marco6: See the other comment.

internal/bindings/server.go
marco6: See the other comment.

app/options.go
letFunny: I had assumed that if you update the go driver that can also dictate changes on previous versions such as deprecations because you _are_ updating the driver. To make that simplifies a bit the story.

letFunny (commented):
Discussed offline, I don't have a strong opinion and I see the alternative of making sure the new version of the driver continues to work with disk mode in the releases that have it, so I am approving. Thanks Marco!

But please let's add some smoke tests with 1.17 and 1.18 and master to make sure linking works at least and dqlite can be invoked.

just-now (approved):

PR #390, opened 2025-08-20
chore: fix tests for bind errors in v1.18

marco6:
This PR fixes the error checking for binding in version 1.18.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -73,7 +73,7 @@ jobs:
 
     - uses: actions/upload-artifact@v4
       with:
-        name: dqlite-benchmark-${{ matrix.os }}-${{ matrix.go }}
+        name: dqlite-benchmark-${{ matrix.os }}-${{ matrix.go }}-${{ matrix.version }}
         path: /tmp/dqlite-benchmark/127.0.0.1:9001/results/*
 
   finish:
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -12,9 +12,10 @@ jobs:
           - '1.13'
           - stable
         os:
-          - ubuntu-20.04
           - ubuntu-22.04
           - ubuntu-24.04
+          - ubuntu-22.04-arm
+          - ubuntu-24.04-arm
         version:
           - libdqlite1.17
           - libdqlite
--- a/driver/integration_test.go
+++ b/driver/integration_test.go
@@ -101,7 +101,7 @@ func TestIntegration_ConstraintError(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


marco6:
The changes in the pipeline will be overwritten by #386 .

driver/integration_test.go
```diff
@@ -112,20 +112,32 @@ func TestIntegration_ExecBindError(t *testing.T) {
...
 	require.NoError(t, err)
 
-	_, err = db.ExecContext(ctx, "INSERT INTO test(n) VALUES(1)", 1)
-	assert.EqualError(t, err, "bind parameters")
+	stmt, err := db.PrepareContext(ctx, "INSERT INTO test(n) VALUES(1)")
+	assert.NoError(t, err)
+
+	_, err = stmt.ExecContext(ctx, 1)
+	require.Error(t, err)
+	if err.Error() != "bind parameters" && err.Error() != "sql: expected 0 arguments, got 1" {
```
letFunny, line 120: Why 2 possible errors?

driver/integration_test.go
```diff
@@ -112,20 +112,32 @@ func TestIntegration_ExecBindError(t *testing.T) {
...
 	defer cancel()
 
-	_, err := db.QueryContext(ctx, "SELECT 1", 1)
-	assert.EqualError(t, err, "bind parameters")
+	stmt, err := db.PrepareContext(ctx, "SELECT 1")
+	assert.NoError(t, err)
+
+	_, err = stmt.QueryContext(ctx, 1)
+	require.Error(t, err)
+	if err.Error() != "bind parameters" && err.Error() != "sql: expected 0 arguments, got 1" {
```
letFunny, line 138: Same comment here.

.github/workflows/build-and-test.yml
```diff
@@ -12,9 +12,10 @@ jobs:
           - '1.13'
           - stable
         os:
-          - ubuntu-20.04
```
letFunny, line 15: So for 1.18, 20.04 is not supported and that is why it is not tested?

letFunny (commented):
Thanks for this Marco, a couple questions and LGTM

driver/integration_test.go
marco6: The old code (1.17) just returned "bind parameters". The new code returns the error from sqlite which is a bit more detailed.

.github/workflows/build-and-test.yml
marco6: The problem is the usual thing that `ubuntu-20.04` does not have runners anymore. As I wrote in the comment above #386 will overwrite this and use a container in this case.

letFunny (approved):

PR #396, opened 2025-08-21
Modernize

simondeziel:
Goes hand in hand with https://github.com/canonical/dqlite-ppa/pull/10 now that Jammy/22.04 and newer have a `golang-1.23-go` package available.

Dropping Focal/22.04 should be fine as it's already no longer officially supported per: https://discourse.canonical.com/t/dqlite-bringing-back-good-habits/6169#p-24333-dropping-support-for-focal-fossa

Superseds https://github.com/canonical/go-dqlite/pull/395


marco6:
Hi! Unfortunately we are still maintaining this project compatible with 20.04: despite dqlite1.18.x does not support it, this Go wrapper is still capable (and used) to wrap dqlite1.17 which in LTS mode.

I will keep this open in case I'm wrong, but I think we cannot easily get rid of Go 1.13 without making a lot of people unhappy.

simondeziel:
> Hi! Unfortunately we are still maintaining this project compatible with 20.04: despite dqlite1.18.x does not support it, this Go wrapper is still capable (and used) to wrap dqlite1.17 which in LTS mode.
> 
> I will keep this open in case I'm wrong, but I think we cannot easily get rid of Go 1.13 without making a lot of people unhappy.

I just found out that 20.04 also comes with `golang-1.18` so maybe that'd be a middle ground.

PR #401, opened 2025-09-04
chore: PPA publishing improvements

marco6:
This PR moves the debian packaging rules here from canonical/dqlite-ppa and changes the trigger so that only new releases trigger the build.


.github/workflows/packages.yml
```diff
@@ -1,11 +1,13 @@
...
 on:
-  push:
-    branches:
-      - v3
+  release:
+    types: [published]
+    branches: [v3]
 jobs:
   build:
-    if: github.repository == 'canonical/go-dqlite'
```
letFunny, line 8: why are you removing this? I thought this line was here to prevent running this workflow and failing

letFunny (approved):

.github/workflows/packages.yml
marco6: This only runs on the  `v3` branch and in case of a release. Also, it runs only on a `ppa` environment which won't (clearly) be available in forks.

So, either someone really forked the project and re-set up the `ppa` env (with the key) and is actively publishing new versions or this can't run.

PR #404, opened 2025-09-07
Modernize bis

simondeziel:
Goes hand in hand with https://github.com/canonical/dqlite-ppa/pull/13 now that Focal/20.04 and newer have a golang-1.22-go package available.

This is a reworked version of #396 but that keeps the same OS compatibility (Focal and up).

Also superseds https://github.com/canonical/go-dqlite/pull/395 (`mattn/go-sqlite3` update).


app/app_test.go
```diff
@@ -1288,7 +1287,7 @@ var appIndex int
 func newDir(t *testing.T) (string, func()) {
 	t.Helper()
 
-	dir, err := ioutil.TempDir("", "dqlite-app-test-")
+	dir, err := os.MkdirTemp("", "dqlite-app-test-")
```
marco6, line 1290: While modernizing I think we should get rid of this function altogether. Go now has [`(*T).TempDir`](https://pkg.go.dev/testing#T.TempDir) which is the better equivalent of this function.

In this sense, it also has [`(*T).Cleanup`](https://pkg.go.dev/testing#T.Cleanup) which should be used everywhere instead of returning a cleanup function (for example in `newAppWithNoTLS`).

app/roles.go
```diff
@@ -265,7 +265,7 @@ func (c *RolesChanges) list(role client.NodeRole, online bool, domains map[uint6
 	nodes := []client.NodeInfo{}
 	for node, metadata := range c.State {
 		if node.Role == role && metadata != nil == online {
-			if domains == nil || (domains != nil && domains[metadata.FailureDomain]) {
+			if domains == nil || domains[metadata.FailureDomain] {
```
marco6, line 268: This line still has a problem as `metadata` can bi `nil` in case `online` is `false`

benchmark/benchmark_test.go
```diff
@@ -22,7 +21,7 @@ const (
 func bmSetup(t *testing.T, addr string, join []string) (string, *app.App, *sql.DB, func()) {
```
marco6, line 21: Same comment as in the other test. We now have a way for helper functions to register cleanups. In the case of temp dirs, it's even easier. Let's use those features and simplify the flow here.

client/client_test.go
```diff
@@ -167,7 +166,7 @@ func addNode(t *testing.T, cli *client.Client, id uint64) (*dqlite.Node, func())
 func newDir(t *testing.T) (string, func()) {
 	t.Helper()
 
-	dir, err := ioutil.TempDir("", "dqlite-replication-test-")
+	dir, err := os.MkdirTemp("", "dqlite-replication-test-")
```
marco6, line 169: Same as other temp dir stuff.

internal/protocol/config.go
```diff
@@ -32,7 +32,7 @@ func (config Config) RetryStrategies() (strategies []strategy.Strategy) {
 			if attempt > 0 {
 				duration := backoffFunc(attempt)
 				// Duration might be negative in case of integer overflow.
-				if !(0 < duration && duration <= cap) {
+				if 0 >= duration || duration > cap {
```
marco6, line 35: While not a big deal, is there a reason for this change?

internal/protocol/message.go
```diff
@@ -578,13 +578,14 @@ func (r *Rows) Close() error {
 	var err error
 	if !r.message.hasBeenConsumed() {
 		slot := r.message.lastByte()
-		if slot == 0xee {
+		switch slot {
+		case 0xee:
```
marco6, line 582: can we name these value as constants given that we are changing this code?

Something like:

```go
const rowsPartMarker = 0xEE;
const rowsDoneMarker = 0xFF;
```

marco6 (changes requested):
Hi @simondeziel! Thanks for this. Few comments below

marco6:
Also @simondeziel, somehow, don't know why the CLA Check is failing. Could you check why? The other PR went through...

PR #419, opened 2026-01-13
Add role eligibility metadata to roles rebalancing

kadinsayani:
This adds role eligibility hints to go-dqlite and enforces them during role adjustments. Nodes can advertise which roles they are allowed to hold via `AllowedRoles`, and `RolesChanges.Adjust` will demote ineligible voters/standbys while filtering promotion candidates. This enables LXD’s new control plane mode to push eligibility into go‑dqlite’s rebalancing logic so promotion/demotion decisions stay centralized in `RolesChanges.Adjust` instead of being duplicated in LXD. With the new metadata, LXD can mark non‑control‑plane members as spare‑only while control‑plane members remain eligible for voter/standby, keeping enforcement consistent with failure‑domain and weight logic. Includes unit tests for demotion and promotion filtering and godoc updates on metadata fields.

Example usage:

```Go
metadata := &client.NodeMetadata{
	FailureDomain: domains[node.Address],
}

// Default: all roles allowed.
metadata.AllowedRoles = client.RoleMaskAll

if controlPlaneActive {
	roles := memberRoles[node.Address]
	if !slices.Contains(roles, db.ClusterRoleControlPlane) {
    // Non-control-plane members are spare-only when control plane mode is active.
		metadata.AllowedRoles = client.RoleMaskSpare
    }
}
```

### DIFF
--- a/app/roles.go
+++ b/app/roles.go
@@ -120,6 +120,9 @@ func (c *RolesChanges) Handover(id uint64) (client.NodeRole, []client.NodeInfo) 
 		candidates = append(c.list(client.StandBy, true, nil), candidates...)
 	}
 
+	// Enforce role eligibility for the role being handed over.
+	candidates = c.filterCandidatesByRole(candidates, node.Role)
+
 	if len(candidates) == 0 {
 		// No online node available to be promoted.
 		return -1, nil
--- a/app/roles.go
+++ b/app/roles.go
@@ -147,6 +147,37 @@ func (c *RolesChanges) Adjust(leader uint64) (client.NodeRole, []client.NodeInfo
 	offlineVoters := c.list(client.Voter, false, nil)
 	offlineStandbys := c.list(client.StandBy, false, nil)
 
+	for _, node := range onlineVoters {
+		// Don't demote the leader here; callers should handle leader handover separately.
+		// AllowedRoles are intentionally not enforced on the current leader to avoid
+		// destabilizing leadership; handover should be used to move leadership first.
+		if node.ID == leader {
+			continue
+		}
+		if c.roleAllowed(node, client.Voter) {
+			continue
+		}
+		if c.roleAllowed(node, client.StandBy) {
+			return client.StandBy, []client.NodeInfo{node}
+		}
+
+		return client.Spare, []client.NodeInfo{node}
+	}
+
+	for _, node := range onlineStandbys {
+		if c.roleAllowed(node, client.StandBy) {
+			continue
+		}
+		if c.roleAllowed(node, client.Spare) {
+			return client.Spare, []client.NodeInfo{node}
+		}
+		if c.roleAllowed(node, client.Voter) {
+			return client.Voter, []client.NodeInfo{node}
+		}
+
+		return -1, nil
+	}
+
 	domainsWithVoters := c.failureDomains(onlineVoters)
 	allDomains := c.allFailureDomains()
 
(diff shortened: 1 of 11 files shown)


tomponline:
Thanks @kadinsayani 

@marco6 would you mind taking a look at this please as this is something we are looking to implement for this cycle. 

Thanks!

app/roles.go
```diff
@@ -147,6 +147,40 @@ func (c *RolesChanges) Adjust(leader uint64) (client.NodeRole, []client.NodeInfo
 	offlineVoters := c.list(client.Voter, false, nil)
 	offlineStandbys := c.list(client.StandBy, false, nil)
 
+	// Enforce role eligibility before making other adjustments.
```
just-now: I'm not convinced that other users of `go-dqlite` will not suffer from this hunk. If you still want it to be here, you can enable it though the configuration (default is off).

client/client.go
```diff
@@ -265,8 +265,12 @@ func (c *Client) Remove(ctx context.Context, id uint64) error {
 
 // NodeMetadata user-defined node-level metadata.
 type NodeMetadata struct {
+	// FailureDomain identifies a failure domain used during role assignment.
 	FailureDomain uint64
-	Weight        uint64
+	// Weight is used to order candidates within a failure domain (lower is preferred).
+	Weight uint64
+	// AllowedRoles controls which roles the node can assume. Zero means all roles allowed.
+	AllowedRoles RoleMask
```
just-now: I see that not all code paths were updated w.r.t. adapt this change. Look at `client/client.go`:
`func (c *Client) Describe(ctx context.Context) (*NodeMetadata, error)`.

As you want to store additional attributes in this structure, maybe, it's worth making them optional, as other users may not need them.

just-now (commented):
Needs additional discussion

app/roles.go
kadinsayani: Default stays "all roles allowed": `AllowedRoles == 0` is treated as allow‑all, so nodes that don’t set it keep the existing behavior. Only nodes that explicitly set `AllowedRoles` get constrained.

client/client.go
kadinsayani: LXD builds `NodeMetadata` locally rather than using `Describe`. `AllowedRoles` is optional, so existing users will not see a behavior change. Is it ok to leave `Describe` unchanged for now since only callers relying on that API would need it populated?

app/roles.go
just-now: OK, if there's a default value for `AllowedRoles`, how do we enforce the user of `struct client.NodeMetadata` to properly set it?

client/client.go
just-now: > AllowedRoles is optional

Optional values don't need default values to be set (this refers to the previous comment). Therefore `Describe()` needs at least to set AllowedRoles to the default value. 

Alternatively, you may want to define AllowedRoles as optional (for examaple: opt.Option[Roles] from go-optional) it could simplify the wiring.

app/roles.go
kadinsayani: The default is equivalent to not setting it at all. If `AllowedRoles` is not included in `client.NodeMetadata`, all roles are allowed for that node.

client/client.go
kadinsayani: `Describe()` does not need to set `AllowedRoles`. When an `int` field is unset, it defaults to 0.

tomponline: @just-now are you saying that because `Describe()` returns `NodeMetadata` populated from dqlite's current state, but that because `AllowedRoles` isn't populated from dqlite's current state, that we should ensure that its indicated as such from the return value of `Describe()` rather than the default value indicating something which is potentially untrue?

tomponline: If so, the `AllowedRoles` could be made a `*RoleMask` so that it can returned as `nil` to indicate the info is not available.

kadinsayani: I've added support for optional response fields to the schema generator. Additionally, `AllowedRoles` is now a `*RoleMask`. `roleAllowed` returns `true` when metadata is missing, `AllowedRoles` is `nil`, or `*AllowedRoles == 0`.

kadinsayani: @just-now I've also patched `Describe` and added an optional `WithAllowedRoles`.

PR #420, opened 2026-02-18
Export role helpers

kadinsayani:
Exports role helper functions to be used by downstream implementations requiring custom role management.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -378,7 +378,7 @@ func (a *App) Handover(ctx context.Context) error {
 			return fmt.Errorf("cluster servers: %w", err)
 		}
 		changes := a.makeRolesChanges(nodes)
-		voters := changes.list(client.Voter, true, nil)
+		voters := changes.List(client.Voter, true, nil)
 
 		for i, voter := range voters {
 			if voter.Address == a.address {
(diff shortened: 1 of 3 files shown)


app/app.go
```diff
@@ -378,7 +378,7 @@ func (a *App) Handover(ctx context.Context) error {
 			return fmt.Errorf("cluster servers: %w", err)
 		}
 		changes := a.makeRolesChanges(nodes)
-		voters := changes.list(client.Voter, true, nil)
+		voters := changes.List(client.Voter, true, nil)
```
just-now, line 381: You don't need to change the whole code. I'd agree on an additional `app_exports.go` with wrappers enabled by a compile option with `-tags` defaulted to off:

```
//go:build appexportsapi
package app

func (x *X) List() { x.list() }
func Bar() { /* ... */ }
func Foo() { /* ... */ }
```

WDYT?

just-now (commented):
Needs more time to think.

app/app.go
just-now: After a discussion I understood that these functions have been already used for the role management and then hidden.

just-now (approved):
Exposing role management functions back ([pr114](https://github.com/canonical/go-dqlite/pull/114)).

